Issue #110, opened 2017-10-17
Define graph schema with case classes that implement traits


pstutz:
Spark can extract `Dataset` schemas from case classes. If we additionally expect these classes to extend a `Node`/`Relationship` interface, then we might use this as a convenient way to define a graph schema. These case class definitions could also be used to define the graph schema for an existing `Dataset`/`DataFrame` (provided that it has the necessary ID columns).

This is what the readme example might look like:
```scala
case class Person(id: Long, name: String) extends Node
case class Friendship(id: Long, source: Long, target: Long) extends Relationship

val persons = Seq(Person(0, "Alice"), Person(1, "Bob")) 
val friendships = Seq(Friendship(0, 0, 1), Friendship(1, 1, 0))
val graph = CAPSGraph.create(NodeScan(persons), RelationshipScan(friendships))
```

This is what the conversion might look like for object instances. Turning an existing Dataset + case class combination into a node/relationship scan would probably not be much different.
```scala
NodeScan/RelationshipScan.apply[T <: PropertyGraph.Node : TypeTag](
    nodes: Seq[T]
)(implicit caps: CAPSSession): NodeScan = {
  /**
    * Turn the nodes into a Dataset (also offer an API for `DataSet` + case class type that extends node/rel).
    * The Node/Relationship traits identify and guarantee the existence of ID columns that encode the graph structure.
    * Get the relationship type and primary node label from the node/relationship class names
    * (e.g. relationshipTag.getSimpleName.toUpperCase).
    * Turn `extraLabels` (if they exist) into additional node labels. All other fields are turned into properties with the field name as the property key.
    */
}
```

If a case class implements `Node`, then a sequence of it or a matching `Dataset` can be turned into a `NodeScan`. Same for `Relationship`/`RelationshipScan`.
```scala
sealed trait Entity extends Product {
  def id: Long
}

trait Node extends Entity {
  def extraLabels: Seq[String] = Seq.empty
}

trait Relationship extends Entity {
  def source: Long

  def target: Long
}
```

What do you think?

pstutz:
Implemented in https://github.com/opencypher/cypher-for-apache-spark/pull/116

PR #116, opened 2017-10-25
Define schema with Scala classes

pstutz:
Implementation of #110

### DIFF
--- a/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
+++ b/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
@@ -25,6 +25,9 @@ import org.opencypher.caps.api.types.{CTNode, CTRelationship, CypherType}
 import org.opencypher.caps.impl.spark.exception.Raise
 import org.opencypher.caps.impl.spark.{SparkColumn, SparkColumnName}
 import org.opencypher.caps.ir.api.Label
+import org.opencypher.caps.api.util.Annotation
+
+import scala.reflect.runtime.universe.TypeTag
 
 sealed trait GraphScan extends Serializable {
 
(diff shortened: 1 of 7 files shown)


pstutz:
Implementation of #110

src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
```diff
@@ -48,6 +51,15 @@ sealed trait GraphScan extends Serializable {
 }
 
 object GraphScan extends GraphScanCompanion[EmbeddedEntity] {
+  implicit def nodesToScan[E <: Node : TypeTag](
+    nodes: Seq[E])(implicit caps: CAPSSession): NodeScan = {
```
Mats-SX: This linebreak seems a little off

src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
```diff
@@ -48,6 +51,15 @@ sealed trait GraphScan extends Serializable {
...
+    NodeScan(nodes)
+  }
+
+  implicit def relationshipsToScan[E <: Relationship : TypeTag](
+    relationship: Seq[E])(implicit caps: CAPSSession): RelationshipScan = {
```
Mats-SX: Also this one

src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
pstutz: I agree. Oddly enough IntelliJ defaults with our changes seems to like it that way and reformats it that way when I change it.

Maybe we can automate style enforcement the same way we automate the license header formatting.

PR #187, opened 2017-10-31
Use Scalafmt Maven plugin

pstutz:


pstutz:
There's also an IntelliJ plugin that allows to format on save.

pstutz:
The Maven plugin is rather slow. Might make more sense for all of us to use the IntelliJ plugin and enable `format on save`.

caps-core/src/main/scala/org/opencypher/caps/CAPS.scala
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
Mats-SX: I'm not sure, but I think there should be double asterisks for the license comment.

boggle:
I have mixed feelings on this; I occasionally like to layout code for readability not sure this is achievable with this. Is there a comment for locally disabling scalafmt?

caps-core/src/main/scala/org/opencypher/caps/CAPS.scala
thobe: No. No there should not.

pstutz: Both IntelliJ formatting and scalafmt disagreed with our license plugin, so I changed its settings to something that the auto-formatters wouldn't touch.

pstutz:
Yes, there's an escape hatch: https://github.com/scalameta/scalafmt/issues/91 
```
// format: OFF
val identityMatrix = Array(1, 0, 0
                                            0, 1, 0
                                            0, 0, 1)
// format: ON
```

pstutz:
Decision at standup: find Scalafmt configuration that causes minimal changes to the current style.

PR #190, opened 2017-11-02
Move Neo4j connection code to CAPS

s1ck:


caps-core/src/main/scala/org/opencypher/caps/impl/spark/io/neo4j/Neo4jGraphLoader.scala
```diff
@@ -45,6 +46,7 @@ object Neo4jGraphLoader {
     import scala.collection.JavaConverters._
 
     val nodeSchema = nodes.aggregate(Schema.empty)({
+      // TODO: what about nodes without labels?
```
Mats-SX: This will be handled by the schema refactoring PR.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/io/neo4j/external/Executor.scala
```diff
@@ -0,0 +1,153 @@
+/**
```
Mats-SX: With the license header PR merged, this needs to change to single asterisk.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/io/neo4j/external/Neo4j.scala
```diff
@@ -0,0 +1,164 @@
+/**
```
Mats-SX: Same here; should be as easy as `mvn license:format`

Mats-SX:
[Integrate neo4j-spark-connector into CAPS](https://trello.com/c/DPJ5S6kH/75-integrate-neo4j-spark-connector-into-caps)

PR #193, opened 2017-11-02
Treat warnings as errors

pstutz:


pstutz:
Decision at standup: try it out and revert it if the flag should cause issues.

PR #195, opened 2017-11-03
Add support for COLLECT aggregation function

DarthMax:


DarthMax:
[collect()](https://trello.com/c/Aakp8gz8/35-collect)

caps-core/src/main/scala/org/opencypher/caps/impl/spark/RowUtils.scala
```diff
@@ -51,11 +51,16 @@ object RowUtils {
       case _: CTNode => (in) => cypherInteger(in.asInstanceOf[Long])
       // TODO: This supports var-expand where we only track rel ids, but it's not right
       case _: CTRelationship => (in) => cypherInteger(in.asInstanceOf[Long])
+      case CTNull => null
```
Mats-SX: This should probably not be a `null` function, but a function that always returns `null`, such as `_ => null`

Mats-SX: Or maybe this case is not used at all, since we call `.material` in the calls

caps-core/src/main/scala/org/opencypher/caps/impl/spark/RowUtils.scala
```diff
@@ -51,11 +51,16 @@ object RowUtils {
...
       case l: CTList => (in) => {
         val converted = in.asInstanceOf[Seq[_]].map(typeToValue(l.elementType))
-
         cypherList(converted.toIndexedSeq)
       }
+      case r: NullableDefiniteCypherType => (in) =>
+        Option(in) match {
+          case Some(v) => typeToValue(r.material)(v)
+          case None => null
+        }
```
Mats-SX: Alternatively, we could skip constructing the option by using for example:

```
(in) => in match {
  case null => null
  case v => typeToValue(r.material)(v)
}
```

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/AggregationAcceptanceTest.scala
```diff
@@ -97,7 +97,7 @@ class AggregationAcceptanceTest extends CAPSTestSuite {
   }
 
   ignore("avg(prop) with only null values in WITH") {
-    val graph = TestGraph("({val:null}),(),()")
+    val graph = TestGraph("({val:NULL}),(),()")
```
Mats-SX: Why is null upper-case?

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/AggregationAcceptanceTest.scala
```diff
@@ -478,6 +478,71 @@ class AggregationAcceptanceTest extends CAPSTestSuite {
...
+  // COLLECT
+  //--------------------------------------------------------------------------------------------------------------------
+
+  test("collect(prop) with integers in WITH") {
+    val graph = TestGraph("({val:2L}),({val:4L}),({val:6L})")
+
+    val result = graph.cypher("MATCH (n) WITH COLLECT(n.val) AS res RETURN res")
+
+    result.records.toMaps should equal(Bag(
+      CypherMap("res" -> Seq(2, 4, 6))
```
Mats-SX: Order is not guaranteed for `.collect()` -- if the implementation changes, this test could break.

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/AggregationAcceptanceTest.scala
DarthMax: this is GDL thing. It only understands uppercase `NULL`

Mats-SX: Ah, makes sense!

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/AggregationAcceptanceTest.scala
```diff
@@ -478,6 +478,71 @@ class AggregationAcceptanceTest extends CAPSTestSuite {
...
   }
 
+  //--------------------------------------------------------------------------------------------------------------------
+  // COLLECT
+  //--------------------------------------------------------------------------------------------------------------------
+
+  test("collect(prop) with integers in WITH") {
+    val graph = TestGraph("({val:2L}),({val:4L}),({val:6L})")
+
+    val result = graph.cypher("MATCH (n) WITH COLLECT(n.val) AS res RETURN res")
```
boggle: Just for reference: function calls like `collect` are written lower case according to the Cypher style guide

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/AggregationAcceptanceTest.scala
boggle: Unless there is an ORDER BY in the WITH, right?

Mats-SX: Correct. It's not put in writing anywhere AFAIK, but that's what I believe we've agreed.

PR #200, opened 2017-11-07
Make logical optimizer more modular

pstutz:


s1ck:
I like that :)

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalOptimizer.scala
```diff
@@ -18,115 +18,123 @@ package org.opencypher.caps.impl.logical
...
+}
 
-    val labelMap = extractLabels(input).groupBy(_._1).mapValues(_.map(_._2))
-
-    def rewrite(root: LogicalOperator): LogicalOperator = {
-      root match {
-        case n@NodeScan(node, in) =>
-          val labels = labelMap.getOrElse(node, Set.empty)
-          val labelNames = labels.map(_.name)
+case object DiscardSelectVoid extends LogicalRewriter {
```
boggle: When does this happen in a real query? Any inference of type VOID usually indicates a typing error.

pstutz: It happens when selecting a property on a label that is not in the graph or when a property is not present on a label. Maybe that was not the intended purpose of inferring `CTVoid`, but I thought that it's nice because it covers many cases where we can already know based on the schema that there won't be any results.

Mats-SX: Those are two different cases. When there is a label predicate for a non-existent label, then yes there will be no records passing through that filter and we know statically that the cardinality is zero.

But when there is a property expression for a non-existent property key, that only affects the value of a field, not the existence of a record. The value of the field should be `null` but the record is still present as usual.

pstutz: I tried to not trigger the optimization in cases where ‘null‘ is permitted by also checking that the type is material: ‘t == CTVoid && t.isMaterial‘. Is that reliable?

Mats-SX: I suppose that makes sense. The only case I can come up with that this would apply to is a (top-level) predicate that statically evaluates to `false` (which `:Foo` does if `Foo` is a non-existent label).

pstutz: Thanks a lot for drawing attention to this. It seems that the typer always returned type CTVoid for properties that it had no information on. We now change this to CTNull and remove the CTVoid optimisation.

Mats-SX: That's great!

PR #202, opened 2017-11-07
Add operators for caching of intermediate results

DarthMax:
- Add CacheStore and CacheRead operators
- Optional Match caches mandatory pattern plans

### DIFF
--- a/caps-core/src/main/scala/org/opencypher/caps/api/spark/CAPSSession.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/api/spark/CAPSSession.scala
@@ -205,7 +205,8 @@ sealed class CAPSSession private(val sparkSession: SparkSession,
       println(physicalPlan.pretty())
     }
 
-    CAPSResultBuilder.from(physicalPlan, logicalPlan)(RuntimeContext(physicalPlannerContext.parameters, optGraphAt))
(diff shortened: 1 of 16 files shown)


DarthMax:
@Mats-SX I general I totally agree with you. The problem I encountered is that, in the physical planning phase, the information about shared plans is already lost. For instance, in the case of the optional match, finding out which part of the plan is optional and which is mandatory would take quite some effort in the physical layer. Basically it would involve traversing the optional plan and finding mandatory plan hidden inside. Which is doable, but I thought I would start with the path of least resistance

Mats-SX:
@DarthMax I was thinking that perhaps you would be able to find that out as you match the `Optional` operator:

```
      case flat.Optional(lhs, rhs, lhsHeader, rhsHeader) =>
        Optional(lhsHeader, rhsHeader, process(lhs), process(rhs))
```

Perhaps we could do `process(lhs)`, plan a `CacheStore`, then do `process(rhs, readFromCache = true, cacheKey)` -- at least it could be worth considering.

DarthMax:
Yes I had that in mind as well. But there is one more problem. In theory the logical optimizer could have rewritten the lhs plan differently than the rhs plan in which case I could not match them any more.

PR #204, opened 2017-11-09
Apply Scalafmt to codbase once

pstutz:
Based on https://github.com/opencypher/cypher-for-apache-spark/pull/187

I don't think we should necessarily use this Maven plugin. I do think it might make sense to consistently format the codebase once and then all locally use the Scalafmt IntelliJ plugin with formatting on save enabled.

I configured Scalafmt to use the IntelliJ style and our default column width.


pstutz:
The formatting changes even turned up an unreachable code warning. :)

caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala
```diff
@@ -203,9 +200,10 @@ object EmbeddedRelationship extends EmbeddedRelationshipBuilder((), (), (), ())
     EmbeddedRelationshipBuilder(entitySlotAndIdSlot, (), (), ())
 }
 
-sealed case class EmbeddedRelationshipBuilder[FROM, VIA, TYP, TO](
-  entitySlotAndIdSlot: VIA, fromSlot: FROM, toSlot: TO, relTypeOrSlotName: TYP
-) {
+sealed case class EmbeddedRelationshipBuilder[FROM, VIA, TYP, TO](entitySlotAndIdSlot: VIA,
```
DarthMax: I really dislike this indention style. It shifts the focus to a very strange place some how.

caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
```diff
@@ -132,28 +136,33 @@ object NodeScan extends GraphScanCompanion[EmbeddedNode] {
...
+          case (schema, label) =>
+            schema.withImpliedLabel(label)
+        }
+        nodeProperties.foldLeft(withLabels) {
+          case (schema, nodeProperty) =>
+            schema.withPropertyKey(nodeProperty)
+        }
       }
-    }.from(nodeRecords)
+      .from(nodeRecords)
```
DarthMax: That looks somewhat odd. On the other hand it is more concise with the rest.

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -151,69 +156,72 @@ case class LabelCombinations(combos: Set[Set[String]]) {
...
-  relKeyMap: PropertyKeyMap,
-  /**
-    * Implied labels for each existing label
-    */
-  impliedLabels: ImpliedLabels,
-  /**
-    * Groups of labels where each group contains possible label combinations.
-    */
-  labelCombinations: LabelCombinations) extends Verifiable {
+                        /**
```
DarthMax: Here the indention thing becomes even clearer. IMHO that's quite a waste of space

caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala
Mats-SX: I agree. I would much prefer a two-space indentation on a new line for multi-line constructors.

```
class Foo(
  s: String,
  i: Int
) {
  // body
}

Comment (pstutz) [reply caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala]: I think this can be changed by setting [`align = none`](http://scalameta.org/scalafmt/#align=none).

File: caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala (hunk 2 above)
Comment (pstutz) [reply caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala]: It dangles a bit, but it really is more consistent. One could argue that it's easier to overlook if it's appended to the last line.

File: caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala (hunk 3 above)
Comment (pstutz) [reply caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala]: I think that `align = none` might fix that one as well.

File: caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala (hunk 1 above)
Comment (pstutz) [reply caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala]: @Mats-SX I think that style is enabled with `align.openParenCallSite = false` instead, so maybe `align = some` would still work.

File: caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala (hunk 2 above)
Comment (pstutz) [reply caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala]: I found a discussion and flag (`includeCurlyBraceInSelectChains = false`) to disable it: https://github.com/scalameta/scalafmt/issues/444#issuecomment-255215698

File: caps-core/src/main/scala/org/opencypher/caps/api/record/RecordHeader.scala [hunk 4]
```diff
@@ -101,52 +101,56 @@ final case class RecordHeader(internalHeader: InternalHeader) {
...
-      v.cypherType match {
-        case t: CTRelationship if targetTypes.isEmpty || t.types.isEmpty => true
-        case CTRelationship(types) =>
-          types.exists(targetTypes.contains)
-        case _ => false
+    slots
+      .collect {
+        case RecordSlot(_, OpaqueField(v)) => v
+      }
+      .filter { v =>
```
Mats-SX: I suppose this is what Max commented on above; not super elegant.

caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
pstutz: I theory `includeCurlyBraceInSelectChains = false` should have fixed this, but in practice it didn't. I assume a future version of Scalafmt will fix that bug.

pstutz:
Based on your comments I force pushed the formatting with most alignments disabled for review.

caps-core/src/main/scala/org/opencypher/caps/api/record/GraphScan.scala
pstutz: Bug has already been reported: https://github.com/scalameta/scalafmt/issues/1041

caps-core/src/main/scala/org/opencypher/caps/api/record/EmbeddedEntity.scala
```diff
@@ -66,10 +69,10 @@ sealed trait VerifiedEmbeddedEntity[V <: EmbeddedEntity] extends Verified[V] {
 }
 
 final case class EmbeddedNode(
-  entitySlot: String,
-  idSlot: String,
-  labelsFromSlotOrImplied: Map[String, Option[String]] = Map.empty,
-  propertiesFromSlots: Map[String, Set[String]] = Map.empty
+    entitySlot: String,
```
Mats-SX: I guess four spaces is OK

pstutz:
Just using Scalafmt.conf with settings above instead, no global reformatting.

PR #206, opened 2017-11-10
Enable tests for unlabeled entities

### LINKED ISSUE #143: Support unlabeled nodes in GDL
https://github.com/opencypher/cypher-for-apache-spark/blob/3afbaabb55d6ec808e7b181f80911e332fef8aa1/src/test/scala/org/opencypher/caps/impl/spark/cypher/FunctionsAcceptanceTest.scala#L98

https://github.com/opencypher/cypher-for-apache-spark/blob/3afbaabb55d6ec808e7b181f80911e332fef8aa1/src/test/scala/org/opencypher/caps/impl/spark/cypher/FunctionsAcceptanceTest.scala#L149

s1ck:
* reflects GDL changes to disable default labels
* fixes #143

### DIFF
--- a/caps-core/src/main/scala/org/opencypher/caps/api/record/RecordHeader.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/api/record/RecordHeader.scala
@@ -22,6 +22,7 @@ import org.opencypher.caps.api.expr._
 import org.opencypher.caps.api.schema.Schema
 import org.opencypher.caps.api.types.{CTBoolean, CTNode, CTString, CypherType, _}
 import org.opencypher.caps.common.syntax._
+import org.opencypher.caps.demo.Configuration.DefaultLabel
 import org.opencypher.caps.impl.record.InternalHeader
 import org.opencypher.caps.impl.spark.exception.Raise
 import org.opencypher.caps.impl.syntax.header.{addContents, _}
@@ -186,7 +187,7 @@ object RecordHeader {
     val optionalNullableKeys = optionalKeys.map { case (k, v) => k -> v.nullable }
     val allKeys: Seq[(String, Vector[CypherType])] = (impliedKeys ++ optionalNullableKeys).toSeq.map { case (k, v) => k -> Vector(v) }
     val keyGroups: Map[String, Vector[CypherType]] = allKeys.groups[String, Vector[CypherType]]
-    val headerLabels = impliedLabels ++ possibleLabels
+    val headerLabels = impliedLabels ++ possibleLabels - DefaultLabel.get()
     val labelHeaderContents = headerLabels.map {
       labelName => ProjectedExpr(HasLabel(node, Label(labelName))(CTBoolean))
     }.toSeq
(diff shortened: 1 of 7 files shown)


s1ck:
@Mats-SX @boggle when we use no label (i.e. empty collections), the SchemaTyper is not able to infer the correct types for properties anymore, since the `Schema` assumes, that `node|relKeys` are associated with a label. This worked before, since there was always a default label. Imho, the system is currently not able to handle non-existing labels. Do you agree?
One option would be to use the empty string for non-existent label and work around that in the udf for the `labels(node)` and `type(edge)` functions. wdyt?

s1ck:
This is not limited to `TestGraph`. The same problem occurs with `NodeScan` e.g.

Mats-SX:
@s1ck You're probably right, and the problem is with the schema here I believe. I agree that we can use the empty string `""` for that, and I did so in the Schema PR when I fixed Neo4j schema inference for unlabeled nodes (this does also not currently work properly).

s1ck:
@Mats-SX I reintroduced the default labels and ignore it in the `labels(n)` function. However, this feels a little inconsistent. For example, in that test:

caps-core/src/main/scala/org/opencypher/caps/impl/spark/Udfs.scala
```diff
@@ -40,7 +40,7 @@ object Udfs {
   def getNodeLabels(labelNames: Seq[Label]): (Any) => Array[String] = {
     case a: mutable.WrappedArray[_] =>
       a.zip(labelNames).collect {
-        case (true, label) => label.name
+        case (true, label) if !label.name.isEmpty => label.name
```
Mats-SX: This should never be allowed to happen

Mats-SX:
For _nodes_, there's no need at all for a default label. It could be convenient for relationships, but only when writing tests. These considerations should in my opinion never leak into the real implementation; this is a test-framework discussion.

caps-core/src/main/scala/org/opencypher/caps/api/record/RecordHeader.scala
```diff
@@ -186,7 +187,7 @@ object RecordHeader {
     val optionalNullableKeys = optionalKeys.map { case (k, v) => k -> v.nullable }
     val allKeys: Seq[(String, Vector[CypherType])] = (impliedKeys ++ optionalNullableKeys).toSeq.map { case (k, v) => k -> Vector(v) }
     val keyGroups: Map[String, Vector[CypherType]] = allKeys.groups[String, Vector[CypherType]]
-    val headerLabels = impliedLabels ++ possibleLabels
+    val headerLabels = impliedLabels ++ possibleLabels - DefaultLabel.get()
```
Mats-SX: Yes, this seems sensible. I would perhaps even hinder this earlier, so that the fake empty label doesn't show up in either `possibleLabels` or `impliedLabels`, but that isn't necessary now; can be done when we get back into remodelling the schema.

PR #208, opened 2017-11-10
Add error handling for list typing issues

### LINKED ISSUE #167: Add error handling for list typing issues
https://github.com/opencypher/cypher-for-apache-spark/blob/3afbaabb55d6ec808e7b181f80911e332fef8aa1/src/main/scala/org/opencypher/caps/impl/typer/SchemaTyper.scala#L456

DarthMax:
fixes #167


caps-core/src/main/scala/org/opencypher/caps/impl/typer/SchemaTyper.scala
```diff
@@ -460,8 +461,7 @@ object SchemaTyper {
           case (CTList(CTInteger), CTInteger) => left
           case (CTList(CTFloat), CTInteger) => CTList(CTNumber)
           case (CTList(CTVoid), _) => CTList(right)
-          // TODO: Throw type error instead
-          case _ => CTVoid
+          case _ => Raise.typeInferenceFailed(s"Could not infer common type of $left and $right")
```
Mats-SX: Since this is in the monad, I think it would be more appropriate to signal this error through the monad's error functionality; you'll find `error()` and `wrong()` calls in other cases in this class.

Mats-SX: If we don't do it that way, we'll stop at this exception instead of keep trying to type other elements and collect all errors at once.

DarthMax: Right, I did not realize that there is a separate error handling in place. I do not think that I fully understand the mechanism, but I have pushed another approach :)

caps-core/src/main/scala/org/opencypher/caps/impl/typer/SchemaTyper.scala
```diff
@@ -449,6 +449,7 @@ object SchemaTyper {
           case _ => None
         }
       }) match {
+        case Some(CTVoid) => wrong[R, TyperError](UnsupportedExpr(expr)) >> pure(Set(Seq(left, right) -> CTVoid))
```
Mats-SX: Yeah, looks good!

PR #213, opened 2017-11-15
Documentation clarifications

moxious:
I'm at the stage of getting "hello world" running with CAPS, which I did -- just adding some notes and clarification that made it easier along the way for people cold-starting like me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,9 @@ target/
 dependency-reduced-pom.xml
 version.txt
 build/test/
+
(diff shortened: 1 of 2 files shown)


Mats-SX:
Hello @moxious and thanks for your PR!

We're currently in waiting for getting a CLA formulated by our legal partners, but once that's done we'll be ready to merge incoming PRs. We'll let this one wait for a bit until we get that sorted.

PR #215, opened 2017-11-17
Migrate to new tree representation

pstutz:
Sorry, this is quite huge change. We didn't want to merge the new node representation before a review and most other changes build on top of that or would interfere with it.

Migrates the flat/physical planners to the new tree representation. Builds on https://github.com/opencypher/cypher-for-apache-spark/pull/212

Summary of changes (including https://github.com/opencypher/cypher-for-apache-spark/pull/212):
- Add Catalyst-style tree node representation that allows elegant top-down and bottom-up rewrites
- Migrate physical planner to new node representation
- Implement physical optimiser that places caching operators using the new node representation and its rewrite support
- Migrate flat and logical planners to new tree node representation
- Change `Expr` and `IRField` to include their `CypherType` in the first parameter list but still exclude it from `equals`/`hashCode`
- Add `asCode` for convenient test case creation: It turns most of the representations we use into strings that can be used to create an instance of the value. In order to make this work without much extra magic `Expr/IRField.cypherType` was moved to the first argument list.
- Add `equalWithTracing` as a replacement for the ScalaTest matcher, because the normal matcher seems to sometimes crash when matching our trees
- Add Scalafmt config that is used by the Scalafmt IntelliJ plugin and incorporates the feedback from https://github.com/opencypher/cypher-for-apache-spark/pull/204. The formatting rules are not applied to the whole codebase.
- Make the flat and physical plans accessible from the query result (`private[caps]`). This allows to more conveniently verify the correctness of plans in tests.


caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/UnaryOperators.scala
```diff
@@ -83,11 +98,12 @@ final case class Alias(expr: Expr, alias: Var, header: RecordHeader) extends Una
       CAPSRecords.create(header, newData)(records.caps)
     }
   }
+
 }
 
-final case class Project(expr: Expr, header: RecordHeader) extends UnaryPhysicalOperator {
+final case class Project(in: PhysicalOperator, expr: Expr, header: RecordHeader) extends UnaryPhysicalOperator {
```
Mats-SX: This is confusing to me; when did this (and some others) lose its child so that it would be re-added by this PR? Surely `Project` always had an `in`?

caps-core/src/main/scala/org/opencypher/caps/ir/api/IRElement.scala
```diff
@@ -19,24 +19,28 @@ import java.net.URI
 
 import org.opencypher.caps.api.schema.Schema
 import org.opencypher.caps.api.types._
+import org.opencypher.caps.impl.common.ModifiedProduct
 import org.opencypher.caps.ir.api.pattern._
 
-trait IRElement {
+trait IRElement extends Product {
```
Mats-SX: Why is this extension necessary?

caps-core/src/main/scala/org/opencypher/caps/ir/api/Token.scala
```diff
@@ -15,13 +15,10 @@
...
   def name: String
 }
 
-final case class Label(name: String) extends AnyVal with Token
-final case class PropertyKey(name: String) extends AnyVal with Token
-final case class RelType(name: String) extends AnyVal with Token
-
-
-
+final case class Label(name: String) extends Token
```
Mats-SX: Why were these changed away from `AnyVal`s? I thought `AnyVal` was a good compiler hint?

caps-core/src/main/scala/org/opencypher/caps/impl/common/TreeNode.scala
```diff
@@ -0,0 +1,159 @@
...
+  def childrenAsSet: Set[T] = children.toSet
+
+  // Optimization: Cache hash code, speeds up repeated computations over high trees.
+  override lazy val hashCode: Int = MurmurHash3.productHash(self)
+
+  def arity: Int = children.length
+
+  def isLeaf: Boolean = height == 1
+
+  def isInner: Boolean = height > 1
```
Mats-SX: What is the intuition behind this name? It seems to me to be `!isLeaf`, which again is the same as `children.nonEmpty`. I would rather call this something like `hasChildren` or `notLeaf` or skip it and use `!isLeaf` in call sites.

caps-core/src/test/scala/org/opencypher/caps/impl/common/TreeNodeTest.scala
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    val expected = Add(Noop(Number(5)), Add(Noop(Number(4)), Noop(Number(3))))
+    println(calculation.asCode)
+
+    val down = calculation.transformDown(addNoops)
+    down should equal(expected)
+
+    val up = calculation.transformUp(addNoops)
+    up should structurallyEqual(expected)
```
Mats-SX: Why is this not actually equal?

caps-core/src/test/scala/org/opencypher/caps/impl/common/TreeNodeTest.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  abstract class Expr extends AbstractTreeNode[Expr] {
```
Mats-SX: I realise this is just a small test, but I would avoid using the common name `Expr` anyway; could be confusing later on. Perhaps `TestExpr` or `CalcExpr` or `Calc`?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalOptimizerTest.scala
```diff
@@ -30,6 +30,15 @@ class LogicalOptimizerTest extends IrTestSuite {
   val logicalGraph = LogicalExternalGraph(testGraph.name, null, Schema.empty)
   val schema = Schema.empty
 
+//  //Helper to create nicer expected results with `asCode`
+//  implicit val specialMappings = Map[Any, String](
+//    schema -> "schema",
+//    emptySqm -> "emptySqm",
+//    logicalGraph -> "logicalGraph",
+//    emptySqm -> "emptySqm",
+//    (CTNode: CTNode) -> "CTNode"
```
Mats-SX: Is this still useful, and is it in the correct place?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
+                Var("a", CTNode),
+                Var("r", CTRelationship),
+                Var("g", CTNode),
+                NodeScan(
+                  Var("a", CTNode),
+                  SetSourceGraph(
+                    LogicalExternalGraph(
+                      "test",
+                      uri,
+                      Schema(
```
Mats-SX: This is the `schema` variable from above; it could be reused instead of recomputed.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
+                        PropertyKeyMap(Map(), Set()),
+                        ImpliedLabels(Map()),
+                        LabelCombinations(Set())
+                      )
+                    ),
+                    Start(
+                      LogicalExternalGraph(
+                        "test",
+                        uri,
+                        Schema(
```
Mats-SX: Same here.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
+                  ),
+                  SolvedQueryModel(Set(nodeA), Set(), Set())
+                ),
+                NodeScan(
+                  Var("g", CTNode),
+                  Start(
+                    LogicalExternalGraph(
+                      "test",
+                      uri,
+                      Schema(
```
Mats-SX: Same

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
+        Not(Equals(Param("p1", CTInteger), Param("p2", CTBoolean), CTBoolean), CTBoolean),
+        NodeScan(
+          Var("a", CTNode),
+          SetSourceGraph(
+            LogicalExternalGraph("test", uri, Schema.empty),
+            Start(LogicalExternalGraph("test", uri, Schema.empty), Set(), emptySqm),
+            emptySqm),
+          SolvedQueryModel(Set(nodeA), Set(), Set())
+        ),
+        SolvedQueryModel(
```
Mats-SX: If we're explicitly stating the `SolvedQueryModel` here, perhaps we actually _should_ consider it for equality in the `LogicalOperator`s.

Like this it is a bit confusing; why state it explicitly like this if it's not part of the check?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
+    result should structurallyEqual(expected)
   }
 
   private val planner = new LogicalPlanner(new LogicalOperatorProducer)
 
   private def plan(ir: CypherQuery[Expr], schema: Schema = Schema.empty) =
     planner.process(ir)(LogicalPlannerContext(schema, Set.empty, (_) => FakeGraphSource(schema)))
 
+  // Returns a successful or the first failed match if there is one.
+  private def combine(m: MatchResult*): MatchResult = {
```
Mats-SX: This code could be shared with the `combine()` below.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/structurallyEqual.scala
```diff
@@ -0,0 +1,113 @@
...
+      m
+    } else {
+      MatchResult(false, m.rawFailureMessage + s"\nTRACE: $traceInfo", m.rawNegatedFailureMessage)
+    }
+  }
+
+  val success = MatchResult(true, "", "")
+
+  // Returns a successful or the first failed match if there is one.
+  def combine(m: MatchResult*): MatchResult = {
```
Mats-SX: This could be used by the above copied `combine()` function.

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/MatchAcceptanceTest.scala
```diff
@@ -16,12 +16,31 @@
...
 
 import org.opencypher.caps.api.value.CypherMap
+import org.opencypher.caps.demo.Configuration.PrintLogicalPlan
 import org.opencypher.caps.test.CAPSTestSuite
 
 import scala.collection.Bag
 
 class MatchAcceptanceTest extends CAPSTestSuite {
 
+  test("match a trivial query") {
```
Mats-SX: Was there a code change in this PR that made this run, when it did not previously?

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/OptionalMatchAcceptanceTest.scala
```diff
@@ -16,6 +16,7 @@
 package org.opencypher.caps.impl.spark.cypher
 
 import org.opencypher.caps.api.value.CypherMap
+import org.opencypher.caps.demo.Configuration.PrintPhysicalPlan
```
Mats-SX: Could be removed

caps-core/src/test/scala/org/opencypher/caps/impl/spark/physical/operators/PhysicalOperatorTest.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  val testContents = Seq(TestContent(42, "foo"), TestContent(23, "bar"))
+
+  case class DummyOp(physicalResult: PhysicalResult) extends LeafPhysicalOperator {
+
+    override def executeLeaf()(implicit context: RuntimeContext): PhysicalResult = physicalResult
+  }
+
+  test("cache operator with single input") {
+    val runtimeContext: RuntimeContext = RuntimeContext.empty
```
Mats-SX: Could add `implicit` here to avoid explicit parameter below

caps-core/src/main/scala/org/opencypher/caps/impl/common/AsCode.scala
```diff
@@ -0,0 +1,97 @@
...
+ */
+package org.opencypher.caps.impl.common
+
+import org.opencypher.caps.api.types.CypherType
+
+/**
+  * Returns a string that can be pasted as an object definition for standard case classes,
+  * some other products, collections and objects.
+  */
+object AsCode {
```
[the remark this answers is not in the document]
pstutz: I agree. It's currently used in making TreeNode argument strings look pretty, dropped it for that. Some trees will look uglier now, but that will probably motivate us to make the `toString` of tree node arguments more consistent or to add those arguments to the argument filter.

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
```diff
@@ -16,44 +16,54 @@
...
+  override protected def includedInComparisons(p: Any): Boolean = !p.isInstanceOf[CypherType]
+
   def cypherType: CypherType
 
   def withoutType: String = toString
 
   override def toString = s"$withoutType :: $cypherType"
 }
 
-final case class Param(name: String)(val cypherType: CypherType = CTWildcard) extends Expr {
```
boggle: Why did we not keep cypherType in a separate arg list? I think this way things work really nice.

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
```diff
@@ -66,10 +76,13 @@ sealed abstract class FlatteningOpExpr(_exprs: Set[Expr]) extends Expr with Seri
...
 
   override def equals(obj: scala.Any) = companion.unapply(obj).contains(exprs)
+
   override def hashCode() = exprs.hashCode() + hashPrime
+
   override def toString = s"$productPrefix(${exprs.mkString(", ")})"
 
   protected def companion: FlatteningOpExprCompanion[_]
+
   protected def hashPrime: Int
```
boggle: Why is this needed? Feels rather odd.

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -48,12 +49,23 @@ object PropertyKeyMap {
...
-final case class PropertyKeyMap(m: Map[String, Map[String, CypherType]])(val conflicts: Set[String] = Set.empty) {
+final case class PropertyKeyMap(m: Map[String, Map[String, CypherType]], conflicts: Set[String] = Set.empty)
+  extends ModifiedProduct {
+
+  // Exclude conflicts product parameter from equals and hashCode
+  override protected def includedInComparisons(p: Any): Boolean = {
+    p match {
+      case _: Set[_] => false
+      case _ => true
+    }
```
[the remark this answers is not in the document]
boggle: `Set.unapply(...).isEmpty` works, too.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
```diff
@@ -13,28 +13,27 @@
...
+import org.opencypher.caps.api.spark.{CAPSRecords, CAPSSession}
 import org.opencypher.caps.impl.logical.LogicalExternalGraph
+import org.opencypher.caps.impl.spark.physical.{PhysicalResult, RuntimeContext}
 
-sealed trait LeafPhysicalOperator extends PhysicalOperator {
-  override def execute(inputs: PhysicalResult*)(implicit context: RuntimeContext): PhysicalResult = {
-    require(inputs.length == 0)
-    executeLeaf()
-  }
+private[spark] abstract class LeafPhysicalOperator extends PhysicalOperator {
```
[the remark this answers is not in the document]
pstutz: It indirectly extends `TreeNode`, which has a type parameter with a context bound `ClassTag`. This allows us to match on the node type parameter `T`. Maybe there would be other issues when using a trait, but this is the first one that appears when I try to use a trait instead.

caps-core/src/main/scala/org/opencypher/caps/impl/common/AbstractTreeNode.scala
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+  * Class that implements the `children` and `withNewChildren` methods using reflection when implementing
+  * `TreeNode` with a case class or case object.
+  *
+  * Requirements: All child nodes need to be individual constructor parameters and their order
+  * in children is their oder in the constructor. Every constructor parameter of type `T` is
+  * assumed to be a child node.
+  */
+abstract class AbstractTreeNode[T <: TreeNode[T] : ClassTag] extends TreeNode[T] {
```
boggle: Could this be a trait instead of a class for easier mixin composition in existing hierarchies?

boggle: Have you considered a second implementation of this based on compile time type tags instead?

caps-core/src/main/scala/org/opencypher/caps/impl/common/AbstractTreeNode.scala
```diff
@@ -0,0 +1,96 @@
...
+  self: T =>
+
+  override lazy val children: Seq[T] = productIterator.toVector.collect { case t: T => t }
+
+  /**
+    * Cache children as a set for faster rewrites.
+    */
+  override lazy val childrenAsSet = children.toSet
+
+  override def withNewChildren(newChildren: Seq[T]): T = {
```
boggle: it shouldn't be hard to update this to also work with constructors with multiple argument list (allowing us to keep CypherType in a separate list).

In fact, this is what the neo4j rewriting code does.

caps-core/src/main/scala/org/opencypher/caps/impl/common/AsCode.scala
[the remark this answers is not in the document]
boggle: I find the name not very telling. Perhaps `asScalaLiteralText` or something like it might work better.

caps-core/src/main/scala/org/opencypher/caps/impl/common/AsCode.scala
```diff
@@ -0,0 +1,97 @@
...
+    else {
+      a match {
+        case null         => "null"
+        case s: String    => s""""$s""""
+        case p: Product   => productAsCode(p)
+        case t: Seq[_]    => traversableAsCode(t)
+        case t: Set[_]    => traversableAsCode(t)
+        case t: Map[_, _] => traversableAsCode(t)
+        case b: Boolean   => b.toString
+        case i: Int       => i.toString
```
boggle: We should do all primitives by default. Neo4j may have a good list somewhere in scala java value conversion. Here I'm missing Byte and Short.

boggle: IndexedSeq as well, perhaps even special case for List and Vector as they are so common.

Also for Map, TreeMaps are sorted etc so this match looks like it may be lossy to me.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
```diff
@@ -13,28 +13,27 @@
...
+  override def executeLeaf()(implicit context: RuntimeContext): PhysicalResult =
     PhysicalResult(records, Map(graph.name -> resolve(graph.uri)))
-  }
 }
 
-final case class StartFrom(records: CAPSRecords, graph: LogicalExternalGraph) extends LeafPhysicalOperator {
-  override def executeLeaf()(implicit context: RuntimeContext) = {
-    PhysicalResult(records, Map(graph.name -> resolve(graph.uri)))
-  }
+final case class StartFromUnit(graph: LogicalExternalGraph)(implicit caps: CAPSSession) extends LeafPhysicalOperator {
```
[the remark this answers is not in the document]
pstutz: The reason is that the session is usually passed around implicitly or not at all, because it is picked up from the input records (which leaf operators don't have). The call to `CAPSRecords.unit()` also asks for an implicit session. It would probably make the most sense to put the session into the `RuntimeContext`, but I'm wary of adding more changes to this PR.

caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatPlanner.scala
```diff
@@ -29,65 +29,73 @@ class FlatPlanner extends DirectCompilationStage[LogicalOperator, FlatOperator,
...
-      case logical.NodeScan(node, in) =>
+      case logical.NodeScan(node, in, _) =>
         producer.nodeScan(node, process(in))
 
-      case logical.Project(it, in) =>
+      case logical.Project(it, in, _) =>
         producer.project(it, process(in))
 
-      case logical.ProjectGraph(graph, in) =>
+      case logical.ProjectGraph(graph, in, _) =>
```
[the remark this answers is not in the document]
boggle: I have half of an argument: This could should be used in the front end where ast nodes have input pos and we're not going to change those. So we need to add support for multiarg lists anyhow. Once we have that, using it for CypherType is the right thing because we can lean on scalac to get haschode and equals right which feels much safer than taking care of equality manually.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/UnaryOperators.scala
pstutz: In a previous design the operator didn't have direct access to its children, but they were passed into its `execute` method.

s1ck: We tested another tree abstraction, which has been merged into master: https://github.com/opencypher/cypher-for-apache-spark/commit/33defbeabff0941575153da5ac980457a019abfc

In that abstraction, the operators had no children information, only the tree nodes.

caps-core/src/main/scala/org/opencypher/caps/ir/api/IRElement.scala
pstutz: Nice find, removed it.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalOptimizer.scala
```diff
@@ -0,0 +1,78 @@
...
+  object InsertCachingOperators extends (PhysicalOperator => PhysicalOperator) {
+    def apply(input: PhysicalOperator): PhysicalOperator = {
+      val replacements = calculateReplacementMap(input)
+        .filterKeys {
+          case _:Start | _:StartFromUnit => false
+          case _ => true
+        }
+
+      val nodesToReplace = replacements.keySet
+      val rule = TreeNode.RewriteRule[PhysicalOperator] {
```
boggle: I'd like us to investigate a bit of implicit magic to not have to write things this explicit. It should be possible to lift partial functions to rewrite rules via implicits and then just write

input.transformDown {
 ...
}

boggle: Also, don't see the need to have RewriteRule a nested class of TreeNode. Feels cumbersome.

caps-core/src/main/scala/org/opencypher/caps/ir/api/IRElement.scala
boggle: If that's really needed, use a self type bound, like `trait IRElement { self: Product => ... }`

caps-core/src/main/scala/org/opencypher/caps/ir/api/Token.scala
boggle: Yes, let's keep this. AnyVal removes boxing and makes things faster! If this stops the machinery from working, we should upgrade the machinery not downgrade the tree :)

pstutz: At some point I ran into a problem with their magic. But when I switch them back now it's all good. Turning them back into AnyVal.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
```diff
@@ -29,173 +29,406 @@ import org.opencypher.caps.impl.logical
...
-  val relR = IRField("r")(CTRelationship)
+  val nodeA = IRField("a", CTNode)
+  val nodeB = IRField("b", CTNode)
+  val nodeG = IRField("g", CTNode)
+  val relR = IRField("r", CTRelationship)
+  val uri = URI.create("test")
+
+  val emptySqm = SolvedQueryModel.empty[Expr]
+
+  // Helper to create nicer expected results with `asCode`
```
boggle: Kill or move to proper place

caps-core/src/test/scala/org/opencypher/caps/impl/common/TreeNodeTest.scala
pstutz: Changed it to ScalaTest normal equal. That equal just fails on some of our larger trees.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalOptimizerTest.scala
pstutz: Yes, it would help update the expected output when the test changes.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
boggle: Hmm wasn't the goal to just carry it around to allow a similar check like in Neo4j one day that asserts that what was supposed to be planned = what was constructed by the planner?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/structurallyEqual.scala
```diff
@@ -0,0 +1,113 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.opencypher.caps.impl.logical
+
+import org.opencypher.caps.impl.common.AsCode._
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+case class structurallyEqual(right: Any) extends Matcher[Any] {
```
boggle: What's the motivation for this? How does it differ from plain old equality?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
pstutz: I moved the solved query model to the end of the first parameter list and it is now also checked by the tests.

caps-core/src/test/scala/org/opencypher/caps/impl/spark/cypher/MatchAcceptanceTest.scala
pstutz: No, I just needed a very minimal match query for testing.

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
pstutz: This wasn't a change made in this PR.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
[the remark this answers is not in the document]
Mats-SX: To me, it makes more sense for the planner to hold the session, and construct the `unit()`, rather than having the plan operator itself tied to the session.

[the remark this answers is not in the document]
pstutz: I agree, should I create an Issue for that or try to address it in this PR?

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
Mats-SX: I suppose we could simplify that by moving it to a `Map[LogicalOperator, SolvedQueryModel]` on `LogicalPlanningContext`.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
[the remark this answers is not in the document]
Mats-SX: I think changing it here makes sense, as it isn't merged yet.

caps-core/src/test/scala/org/opencypher/caps/impl/logical/LogicalPlannerTest.scala
pstutz: Makes sense. Should I create an issue or refactor for this PR?

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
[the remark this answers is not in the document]
pstutz: Just committed it.

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
pstutz: It allows for `.asCode` without a lot of reflective magic or having to add boilerplate code to `Expr` instances.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalOptimizer.scala
pstutz: You're right, it's a good idea to get rid of the rule-wrapper and accept a partial function directly. Still, we're not that far away, this works:
```
    calculation.transformDown { RewriteRule {
        case Add(Number(n1), Number(n2)) => Number(n1 + n2)
      }
    }
```

caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatPlanner.scala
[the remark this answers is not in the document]
pstutz: The main gain is really being able to conveniently create test data with `asCode`. That would not be impossible with the extra parameter list, but it would either require reflection magic to grab the right constructor and all values for it or it would require extra boiler plate code in implementing classes. I agree that the awkwardness of blank matchers at the end and that of dangling empty parentheses for constructors approximately balance each other out.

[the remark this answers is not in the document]
Mats-SX: Well, there's also the conceptual difference. A Scala-aware reader will know immediately that the field will not be part of equality checks from the second parameter list, while it will take a lot more time to gain the same knowledge with the current code.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalPlanner.scala
```diff
@@ -117,7 +117,7 @@ class PhysicalPlanner extends DirectCompilationStage[FlatOperator, PhysicalOpera
           case _                       => Raise.invalidArgument("an external graph", sourceOp.sourceGraph.toString)
         }
 
-        val startFrom = StartFromUnit(externalGraph)
+        val startFrom = StartFromUnit(externalGraph, CAPSRecords.unit())
```
s1ck: Why did you reintroduce the parameter? We ran into an issue with caching when this was created externally.

caps-core/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/LeafOperators.scala
```diff
@@ -13,28 +13,29 @@
...
     PhysicalResult(records, Map(graph.name -> resolve(graph.uri)))
-  }
+
 }
 
-final case class StartFrom(records: CAPSRecords, graph: LogicalExternalGraph) extends LeafPhysicalOperator {
-  override def executeLeaf()(implicit context: RuntimeContext) = {
-    PhysicalResult(records, Map(graph.name -> resolve(graph.uri)))
-  }
+final case class StartFromUnit(graph: LogicalExternalGraph, unitRecord: CAPSRecords) extends LeafPhysicalOperator {
```
s1ck: Why did you reintroduce the parameter? We ran into an issue with caching when this was created externally.

s1ck: The issue is, that relation ship scans in an `ExpandSource` are created within the physical planner based on that `StartFromUnit` operator. If the records are part of the constructor, then subtrees which do the same expansion (with identical variables) as in optional match won't be recognized as equal and hence not be cached.

pstutz: My bad. I was asked to clean it up and had forgotten that it caused that problem. I'll add a test case to detect this and revert.

PR #218, opened 2017-11-22
Schema remodelling

Mats-SX:
Remodels the Schema to more closely resemble the ScanGraph model:

- Each label combination (`Set[String]`) is mapped to a set of typed property keys
- Label-less nodes are mapped using an empty set

From old PR:
- Add `.isEmpty()`
- Prettier `.toString()`
- Proper testing of loading from Neo4j
- Add docs explaining the contract of `.withNodes/RelsPropertyKeys()`


caps-core/src/main/scala/org/opencypher/caps/api/schema/LabelPropertyMap.scala
```diff
@@ -0,0 +1,109 @@
...
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+package org.opencypher.caps.api.schema
+
+import org.opencypher.caps.api.schema.PropertyKeys.PropertyKeys
+import org.opencypher.caps.api.types.CypherType
+import org.opencypher.caps.impl.util.MapUtils._
+
+object PropertyKeys {
```
boggle: Since things are being split up, perhaps this should go in a separate file, too.

DarthMax: Mhh I think that having this in the same class as the LabelPropertyMap, as thy are coupled rather tightly.

caps-core/src/main/scala/org/opencypher/caps/impl/util/MapUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+trait MapUtils {
+
+  def merge[A, B](a: Map[A, B], b: Map[A, B])(mergeValues: (B, B) => B): Map[A, B] =
+    (a.keySet ++ b.keySet)
+      .map(k => (k, a.get(k), b.get(k)))
+      .collect {
+        case (k, Some(v1), Some(v2)) => k -> mergeValues(v1, v2)
+        case (k, Some(v1), _) => k -> v1
+        case (k, _, Some(v2)) => k -> v2
+      }.toMap
```
Mats-SX: Nice! :)

caps-core/src/test/scala/org/opencypher/caps/api/record/NodeScanTest.scala
```diff
@@ -60,14 +60,8 @@ class NodeScanTest extends CAPSTestSuite {
     ))
 
     nodeScan.schema should equal (Schema.empty
-      .withImpliedLabel("A","B")
-      .withImpliedLabel("B","A")
-      .withImpliedLabel("C","A")
-      .withImpliedLabel("C","B")
-      .withLabelCombination("A","C")
-      .withLabelCombination("B","C")
```
Mats-SX: Interesting! This was clearly a bug before; these combinations were not possible for the scan.

caps-core/src/test/scala/org/opencypher/caps/api/record/RecordHeaderTest.scala
```diff
@@ -329,4 +331,130 @@ class RecordHeaderTest extends BaseTestSuite {
...
+      ProjectedExpr(Property(b, PropertyKey("a"))(CTString.nullable)),
+      ProjectedExpr(Property(b, PropertyKey("b"))(CTInteger.nullable)),
+      ProjectedExpr(Property(b, PropertyKey("c"))(CTFloat)),
+      ProjectedExpr(Property(b, PropertyKey("extra"))(CTBoolean.nullable))
+    )))._1)
+
+    cHeader should equal(RecordHeader.empty.update(addContents(Seq(
+      OpaqueField(c),
+      ProjectedExpr(HasLabel(c, Label("C"))(CTBoolean))
+    )))._1)
```
Mats-SX: Good addition!

caps-core/src/test/scala/org/opencypher/caps/api/record/RecordHeaderTest.scala
```diff
@@ -329,4 +331,130 @@ class RecordHeaderTest extends BaseTestSuite {
...
+      OpaqueField(ab),
+      ProjectedExpr(HasLabel(ab, Label("A"))(CTBoolean)),
+      ProjectedExpr(HasLabel(ab, Label("B"))(CTBoolean)),
+      ProjectedExpr(Property(ab, PropertyKey("a"))(CTString)),
+      ProjectedExpr(Property(ab, PropertyKey("b"))(CTInteger.nullable)),
+      ProjectedExpr(Property(ab, PropertyKey("c"))(CTFloat))
+    )))._1)
+  }
+
+  test("node from schema with implication") {
```
Mats-SX: This test doesn't really add anything to the one above, but if/when we choose to optimise away our label columns, this would become interesting.

caps-core/src/test/scala/org/opencypher/caps/api/record/RecordHeaderTest.scala
```diff
@@ -329,4 +331,130 @@ class RecordHeaderTest extends BaseTestSuite {
...
+  }
+
+  test("relationship from schema with given relationship types") {
+    val schema = Schema.empty
+      .withRelationshipPropertyKeys("A")("a" -> CTString, "b" -> CTInteger.nullable)
+      .withRelationshipPropertyKeys("B")("a" -> CTString, "b" -> CTInteger)
+
+    val e = Var("e")(CTRelationship("A"))
+
+    val eHeader = RecordHeader.relationshipFromSchema(e, schema, Set("A", "B"))
```
Mats-SX: Could put both rel-types on the `Var` here.

caps-core/src/test/scala/org/opencypher/caps/api/schema/SchemaTest.scala
```diff
@@ -62,18 +65,32 @@ class SchemaTest extends BaseTestSuite {
...
+    schema.labels should equal(Set("Person", "Employee", "Foo", "Bar", "Dog"))
   }
 
   test("should get chained implications correct") {
-    val schema = Schema.empty.withImpliedLabel("Employee", "Person")
-      .withImpliedLabel("Person", "Human")
-      .withImpliedLabel("Person", "Someone")
+//    val schema = Schema.empty.withImpliedLabel("Employee", "Person")
+//      .withImpliedLabel("Person", "Human")
+//      .withImpliedLabel("Person", "Someone")
```
Mats-SX: Remove

caps-core/src/test/scala/org/opencypher/caps/api/spark/CAPSGraphOperationsTest.scala
```diff
@@ -111,37 +113,37 @@ class CAPSGraphOperationsTest extends CAPSTestSuite {
...
-        Row(/* n */ 200L, /* ____n:Person */ true, /* ____n:Swedish */ false, /* ____n:Programmer */ true, /* ____n:Book */ false, /* ____n_dot_lucky_bar_numberINTEGER */ 23L, /* ____n_dot_nameSTRING */ "Bob", /* ____n_dot_yearINTEGER */ null, /* ____n_dot_languageSTRING */ "D", /* ____n_dot_titleSTRING */ null),
-        Row(/* n */ 300L, /* ____n:Person */ true, /* ____n:Swedish */ false, /* ____n:Programmer */ true, /* ____n:Book */ false, /* ____n_dot_lucky_bar_numberINTEGER */ 84L, /* ____n_dot_nameSTRING */ "Eve", /* ____n_dot_yearINTEGER */ null, /* ____n_dot_languageSTRING */ "F", /* ____n_dot_titleSTRING */ null),
-        Row(/* n */ 400L, /* ____n:Person */ true, /* ____n:Swedish */ false, /* ____n:Programmer */ true, /* ____n:Book */ false, /* ____n_dot_lucky_bar_numberINTEGER */ 49L, /* ____n_dot_nameSTRING */ "Carl", /* ____n_dot_yearINTEGER */ null, /* ____n_dot_languageSTRING */ "R", /* ____n_dot_titleSTRING */ null)
+      Bag(
```
Mats-SX: Perhaps we should also have an assertion on the header?

caps-core/src/test/scala/org/opencypher/caps/api/spark/CAPSGraphOperationsTest.scala
```diff
@@ -111,37 +113,37 @@ class CAPSGraphOperationsTest extends CAPSTestSuite {
...
-        Row(/* ____source(r) */ 2L, /* r */ 5L, /* ____type(r) */ "KNOWS", /* ____target(r) */ 4L, /* ____r_dot_sinceINTEGER */ 2013L, /* ____r_dot_recommendsBOOLEAN */ null),
-        Row(/* ____source(r) */ 3L, /* r */ 6L, /* ____type(r) */ "KNOWS", /* ____target(r) */ 4L, /* ____r_dot_sinceINTEGER */ 2016L, /* ____r_dot_recommendsBOOLEAN */ null),
-        Row(/* ____source(r) */ 100L, /* r */ 100L, /* ____type(r) */ "READS", /* ____target(r) */ 10L, /* ____r_dot_sinceINTEGER */ null, /* ____r_dot_recommendsBOOLEAN */ true),
-        Row(/* ____source(r) */ 200L, /* r */ 200L, /* ____type(r) */ "READS", /* ____target(r) */ 40L, /* ____r_dot_sinceINTEGER */ null, /* ____r_dot_recommendsBOOLEAN */ true),
-        Row(/* ____source(r) */ 300L, /* r */ 300L, /* ____type(r) */ "READS", /* ____target(r) */ 30L, /* ____r_dot_sinceINTEGER */ null, /* ____r_dot_recommendsBOOLEAN */ true),
-        Row(/* ____source(r) */ 400L, /* r */ 400L, /* ____type(r) */ "READS", /* ____target(r) */ 20L, /* ____r_dot_sinceINTEGER */ null, /* ____r_dot_recommendsBOOLEAN */ false)
+      Bag(
```
Mats-SX: Same here

caps-core/src/test/scala/org/opencypher/caps/impl/spark/io/neo4j/Neo4jSchemaLoaderTest.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  test("read nodes") {
+    val graph = Neo4jGraphLoader.loadSchema(neo4jConfig, "MATCH (n) RETURN n", emptyQ)
+
+    graph.schema should equal(
+      Schema.empty
+        .withNodePropertyKeys("Person")("name" -> CTString, "age" -> CTInteger.nullable, "salary" -> CTInteger.nullable, "team" -> CTString.nullable)
+        .withNodePropertyKeys("Employee")("name" -> CTString, "salary" -> CTInteger, "team" -> CTString.nullable)
+        .withNodePropertyKeys("Driver")("name" -> CTString.nullable, "fast" -> CTBoolean.nullable)
+        .withNodePropertyKeys("")("pi" -> CTFloat.nullable)
```
Mats-SX: Err! Fix this. (Related to above comment.) Should never have empty strings like this (perhaps assert that in the call even?)

caps-core/src/test/scala/org/opencypher/caps/impl/util/MapUtilsTest.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import org.opencypher.caps.test.BaseTestSuite
+
+class MapUtilsTest extends BaseTestSuite {
+
+  test("merge") {
+    val map1 = Map(1 -> 1, 2 -> 2)
+    val map2 = Map(1 -> 2, 3 -> 3)
+
+    MapUtils.merge(map1, map2)(_ + _) should equal(Map(1 -> 3, 2 -> 2, 3 -> 3))
```
Mats-SX: This is such a cute function :D

caps-core/src/test/scala/org/opencypher/caps/test/support/DebugOutputSupport.scala
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package org.opencypher.caps.test.support
+
+import org.apache.spark.sql.Row
+
+import scala.collection.Bag
+import scala.collection.immutable.HashedBagConfiguration
+
+trait DebugOutputSupport {
```
Mats-SX: Perhaps we should call this `RowDebugSupport` or `RowsToBagSupport` or smth with `Row` in the name.

caps-core/src/main/scala/org/opencypher/caps/api/schema/LabelPropertyMap.scala
s1ck: @DarthMax `RelTypePropertyMap` is also using `PropertyKeys` and shares the `++` logic with `LabelPropertyMap`. Maybe they should all be in one file sealed file.

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -15,143 +15,48 @@
...
-            Some(s"Conflicting schema for '$classifier'! Key '$k' has type $t but also has type ${newKeys(k)}")
-          case _ =>
-            None
-        }
-    }.flatten.toSet
-    copy(m.updated(classifier, oldKeys ++ newKeys))(conflicts = conflicts ++ newConflicts)
-  }
 
-  def keys = m.values.flatMap(_.keySet).toSet
+  object NoLabel extends Set[String] {
```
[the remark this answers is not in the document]
DarthMax: We needed an abstraction to identify the cases where we specifically want to express any label, i.e all labels including the empty one, and only the empty label (the above case). Without these abstractions we had lots of problems separating these cases.

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -15,143 +15,48 @@
...
-    common.foldLeft(withUniqueRight) {(map, key) => map.updated(key, joinF(left(key), right(key)))}
+    override def iterator: Iterator[String] = Iterator.empty
   }
-}
 
-object ImpliedLabels {
-  val empty: ImpliedLabels = ImpliedLabels(Map.empty)
-}
+  object AllLabels extends Set[String] {
+    override def contains(elem: String): Boolean = false
```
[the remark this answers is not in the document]
DarthMax: that is correct

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -15,143 +15,48 @@
...
-  def transitiveImplicationsFor(known: Set[String]): Set[String] = {
-    val next = known.flatMap(implicationsFor)
-    if (next == known) known else transitiveImplicationsFor(next)
-  }
+    override def -(elem: String): Set[String] = this
 
-  def withImplication(source: String, target: String): ImpliedLabels = {
-    val implied = implicationsFor(source)
-    if (implied(target)) this else copy(m = m.updated(source, implied + target))
+    override def iterator: Iterator[String] = Iterator.empty
```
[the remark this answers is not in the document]
DarthMax: in essence we do not really want to iterate over the elements since we would have to know the schema to do this. The idea was to treat AllLabels and NoLabel allays separately.

caps-core/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -163,91 +68,230 @@ final case class Schema(
...
+  }
+
+  def relationshipKeyType(types: Set[String], key: String): Option[CypherType] = {
+    // relationship types have OR semantics: empty set means all types
+    val relevantTypes = if (types.isEmpty) relationshipTypes else types
+
+    relevantTypes.map(relationshipKeys).foldLeft(CTVoid: CypherType) {
+      case (inferred, next) => inferred.join(next.getOrElse(key, CTNull))
+    } match {
+      case CTNull => None
```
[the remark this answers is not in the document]
DarthMax: right

caps-core/src/main/scala/org/opencypher/caps/impl/typer/SchemaTyper.scala
```diff
@@ -65,41 +67,17 @@ object SchemaTyper {
...
-              if (labels.isEmpty)
-                schema.labels.map(schema.nodeKeys).foldLeft[CypherType](CTVoid) {
-                  case (acc, next) => acc.join(next.getOrElse(name, CTVoid))
-                }
-              else {
-                val keys = labels.map(schema.nodeKeys).reduce(_ ++ _)
-                keys.getOrElse(name, CTVoid)
-              }
-            }
+            val relevantLabels = if (labels.isEmpty) AnyGiven(AllLabels) else AllGiven(labels)
```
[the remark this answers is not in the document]
DarthMax: Yes this in used for the case `Match (n)` so a node with whatever label.

Mats-SX:
This PR had more bugs than I first anticipated. Additional fixes:

- Computing nullable relationship properties from other relationship types
- Sort relationship property slots in RecordHeader computation
- Use empty set when loading unlabeled nodes from Neo4j (not empty string)

Additional fixes necessary: compute correct type for a node property when only partial label information is present. This relates to the `AllGiven(labels)` call currently in `SchemaTyper`; if there are label combinations `:A:B` but not `:A` on its own, then the code path must understand that there may be property keys for `:A` nodes indirectly.

caps-core/src/main/scala/org/opencypher/caps/impl/typer/SchemaTyper.scala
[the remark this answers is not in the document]
Mats-SX: The thing is that here the `labels` set is a _lower bound_, not an exact match. It has been a `Map[String, Boolean]`, in which case it had the power to be exact, but now it is just a set of expected labels, and any unmentioned labels are possible (subject to the schema). The schema needs to understand it as such, which is what `AllGiven` accomplishes here, I think. On the inside of `nodeKeys()` it is interpreted directly, however, so something needs to be done there similar to what we do in `RecordHeader#nodeFromSchema`.

caps-core/src/test/scala/org/opencypher/caps/api/record/RecordHeaderTest.scala
Mats-SX: Scratch that, we're already doing it.

PR #220, opened 2017-11-24
Tree node fixes

pstutz:
Fixes the string representation of trees, adds tests to verify the string representation stays pretty, and slightly improves caching of reflected copy methods by using ThreadLocal instead of a concurrent map.

### DIFF
--- a/caps-core/src/main/scala/org/opencypher/caps/impl/common/AbstractTreeNode.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/impl/common/AbstractTreeNode.scala
@@ -27,7 +27,7 @@ import scala.reflect.ClassTag
   * in children is their order in the constructor. Every constructor parameter of type `T` is
   * assumed to be a child node.
   */
-abstract class AbstractTreeNode[T <: TreeNode[T] : ClassTag] extends TreeNode[T] {
+abstract class AbstractTreeNode[T <: TreeNode[T]: ClassTag] extends TreeNode[T] {
   self: T =>
 
   override lazy val children: Seq[T] = productIterator.toVector.collect { case t: T => t }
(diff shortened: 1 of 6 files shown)


caps-core/src/test/scala/org/opencypher/caps/impl/common/TreeNodeTest.scala
```diff
@@ -44,19 +44,42 @@ class TreeNodeTest extends FunSuite with Matchers {
...
+    Add(Number(1), Number(2)).toString should equal("Add")
+  }
+
+  test("pretty") {
+    calculation.pretty should equal("""|-Add
+· |-Number(5)
+· |-Add
+· · |-Number(4)
+· · |-Number(3)
+""")
```
Mats-SX: I think we can make this prettier by using `.stripMargin` with a custom character, perhaps `!` or `#`.

caps-core/src/main/scala/org/opencypher/caps/impl/common/AbstractTreeNode.scala
```diff
@@ -42,55 +42,62 @@ abstract class AbstractTreeNode[T <: TreeNode[T] : ClassTag] extends TreeNode[T]
...
-    val className = instance.getClass.getName
-    cachedCopyMethods.getOrElse(className, {
-      val instanceMirror = mirror.reflect(instance)
-      val tpe = instanceMirror.symbol.asType.toType
-      val copyMethodSymbol = tpe.decl(TermName("copy")).asMethod
-      val copyMethod = instanceMirror.reflectMethod(copyMethodSymbol)
-      cachedCopyMethods.put(className, copyMethod)
-      copyMethod
-    })
+  private lazy val cachedCopyMethods = new ThreadLocal[Map[Class[_], MethodMirror]]() {
```
Mats-SX: What is the advantage to a `ThreadLocal` vs a `ConcurrentHashMap`?

pstutz: It's more elegant and probably faster, as no real concurrency is needed for this use case. Saw that while researching the Neo4j reflection code and liked the idea.

PR #225, opened 2017-11-30
Add support for basic pattern predicates

s1ck:
This PR introduces support for pattern expressions without additional node and relationship predicates (i.e. such based on properties).

### DIFF
--- a/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
@@ -201,6 +201,16 @@ class FlatOperatorProducer(implicit context: FlatPlannerContext) {
     Optional(lhs, rhs, lhs.header, rhs.header)
   }
 
+  def planPatternPredicate(pattenExpr: PatternExpr, lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
+    val (header, status) = lhs.header.update(addContent(ProjectedField(pattenExpr.predicateField, pattenExpr)))
+
+    status match {
+      case _: Added[_]        => PatternPredicate(pattenExpr.predicateField, lhs, rhs, header)
+      case f: FailedToAdd[_]  => Raise.slotNotAdded(f.toString)
+      case _                  => Raise.impossible("Invalid RecordHeader update status.")
+    }
+  }
+
   def orderBy(sortItems: Seq[SortItem[Expr]], sourceOp: FlatOperator): FlatOperator = {
     OrderBy(sortItems, sourceOp, sourceOp.header)
   }
--- a/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
@@ -198,7 +198,7 @@ class FlatOperatorProducer(implicit context: FlatPlannerContext) {
   }
 
   def planOptional(lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
-    Optional(lhs, rhs, lhs.header, rhs.header)
+    Optional(lhs, rhs, rhs.header)
   }
 
   def planPatternPredicate(pattenExpr: PatternExpr, lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
--- a/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
@@ -201,11 +201,11 @@ class FlatOperatorProducer(implicit context: FlatPlannerContext) {
     Optional(lhs, rhs, rhs.header)
   }
 
-  def planPatternPredicate(pattenExpr: PatternExpr, lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
-    val (header, status) = lhs.header.update(addContent(ProjectedField(pattenExpr.predicateField, pattenExpr)))
+  def planPatternPredicate(patternExpr: PatternExpr, lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
+    val (header, status) = lhs.header.update(addContent(ProjectedField(patternExpr.predicateField, patternExpr)))
 
     status match {
-      case _: Added[_]        => PatternPredicate(pattenExpr.predicateField, lhs, rhs, header)
+      case _: Added[_]        => PatternPredicate(patternExpr.predicateField, lhs, rhs, header)
       case f: FailedToAdd[_]  => Raise.slotNotAdded(f.toString)
       case _                  => Raise.impossible("Invalid RecordHeader update status.")
     }
--- a/caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalOptimizer.scala
+++ b/caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalOptimizer.scala
@@ -18,6 +18,7 @@ package org.opencypher.caps.impl.logical
 import org.opencypher.caps.api.expr.{HasLabel, Var}
 import org.opencypher.caps.api.types.{CTBoolean, CTNode}
 import org.opencypher.caps.impl.DirectCompilationStage
+import org.opencypher.caps.impl.exception.Raise
 import org.opencypher.caps.ir.api.Label
 
 object LogicalOptimizer extends DirectCompilationStage[LogicalOperator, LogicalOperator, LogicalPlannerContext] {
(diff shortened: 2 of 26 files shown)


caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
```diff
@@ -198,7 +198,17 @@ class FlatOperatorProducer(implicit context: FlatPlannerContext) {
   }
 
   def planOptional(lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
-    Optional(lhs, rhs, lhs.header, rhs.header)
+    Optional(lhs, rhs, rhs.header)
+  }
+
+  def planPatternPredicate(pattenExpr: PatternExpr, lhs: FlatOperator, rhs: FlatOperator): FlatOperator = {
```
Mats-SX: misspelled variable: pattenExpr

caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
```diff
@@ -198,7 +198,17 @@ class FlatOperatorProducer(implicit context: FlatPlannerContext) {
...
+    val (header, status) = lhs.header.update(addContent(ProjectedField(pattenExpr.predicateField, pattenExpr)))
+
+    status match {
+      case _: Added[_]        => PatternPredicate(pattenExpr.predicateField, lhs, rhs, header)
+      case f: FailedToAdd[_]  => Raise.slotNotAdded(f.toString)
+      case _                  => Raise.impossible("Invalid RecordHeader update status.")
```
Mats-SX: I guess that if `Found[_]` is an answer, it could mean we already computed the pattern predicate?

```
MATCH (a)
WHERE (a)-->(:Foo) AND (a)-->(:Foo) // we only need to compute once
...
```

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalOptimizer.scala
```diff
@@ -41,9 +42,9 @@ object LogicalOptimizer extends DirectCompilationStage[LogicalOperator, LogicalO
   }
 
   def pushLabelsIntoScans(labelMap: Map[Var, Set[String]]): PartialFunction[LogicalOperator, LogicalOperator] = {
-    case NodeScan(v @ Var(name), in, solved) =>
+    case ns @ NodeScan(v @ Var(name), in, solved) =>
       val updatedLabels = labelMap(v)
-      val updatedVar = Var(name)(CTNode(updatedLabels))
+      val updatedVar = Var(name)(CTNode(ns.labels ++ updatedLabels))
```
Mats-SX: Nice

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
```diff
@@ -31,8 +31,11 @@ import scala.annotation.tailrec
 final case class LogicalPlannerContext(
   ambientGraphSchema: Schema,
   inputRecordFields: Set[Var],
-  resolver: String => GraphSource
-)
+  resolver: String => GraphSource,
+  sourceGraph: IRGraph
```
Mats-SX: Is this like 'current' source graph as we write the logical plan? Like we have in `IRBuilderContext`?

caps-core/src/main/scala/org/opencypher/caps/ir/impl/ExpressionConverter.scala
```diff
@@ -15,17 +15,18 @@
...
 import org.opencypher.caps.api.types._
+import org.opencypher.caps.impl.flat.FreshVariableNamer
 import org.opencypher.caps.impl.parse.RetypingPredicate
 import org.opencypher.caps.ir.api.{Label, PropertyKey, RelType}
 import org.opencypher.caps.ir.impl.FunctionUtils._
 
 import scala.language.implicitConversions
 
-object ExpressionConverter {
+final class ExpressionConverter(patternConverter: PatternConverter = new PatternConverter(Map.empty)) {
```
Mats-SX: Do we need the default parameter value? For tests?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
```diff
@@ -53,62 +53,142 @@ final case class ValueJoin(
...
     val joinedRecords =
-      joinDFs(lhsData, reducedRhsData, rhsHeader, joinCols)("leftouter", deduplicate = true)(left.records.caps)
+      joinDFs(left.records.data, reducedRhsData, header, joinCols)("leftouter", deduplicate = true)(left.records.caps)
 
     PhysicalResult(joinedRecords, left.graphs ++ right.graphs)
   }
+}
+
+/**
+  * This operator performs a left outer join between the mandatory path and the pattern path. If, for a given mandatory
```
Mats-SX: Perhaps we shouldn't say `mandatory match` here, since that concept exists in Cypher and means something specific.

Mats-SX: We could instead talk about 'already matched', or 'expected matched' or 'already bound'.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
```diff
@@ -53,62 +53,142 @@ final case class ValueJoin(
...
+
+    val joinCols = joinColumnMapping.map(t => t._1 -> distinctRightData.col(t._3))
 
+    val joinedRecords =
+      joinDFs(left.records.data, distinctRightData, rhsHeader, joinCols)("leftouter", deduplicate = true)(left.records.caps)
+
+    val rightNonJoinColumns = distinctRightData.columns.toSet -- joinColumnMapping.map(_._3)
+
+    // We only need to check one non-join column for its value to decide if the pattern exists
+    assert(rightNonJoinColumns.nonEmpty)
```
Mats-SX: Assertions are a bit scary in production code...

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
```diff
@@ -53,62 +53,142 @@ final case class ValueJoin(
...
+    val rightNonJoinColumn = distinctRightData.col(rightNonJoinColumns.head)
+
+    // If the non-join column contains no value we set the predicate field to false, otherwise true.
+    // After that we drop all right columns and only keep the predicate field.
+    // The predicate field is later checked by a filter operator.
```
Mats-SX: Nice comments

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/cypher/PredicateAcceptanceTest.scala
```diff
@@ -226,4 +226,148 @@ class PredicateAcceptanceTest extends CAPSTestSuite {
     // And
     result.graphs shouldBe empty
   }
+
+  test("basic pattern predicate") {
+    // Given
+    val given = TestGraph("(v {id:1L})-->({id:2L})-->({id:3L})<--(v)")
```
Mats-SX: Maybe a good idea would be to have another instance of two out-hops that _doesn't_ match the same destination node. Could be just a fourth node from nbr 3 here (two hops from 2 that doesn't end up at 3)

caps-core/src/main/scala/org/opencypher/caps/impl/flat/FlatOperatorProducer.scala
DarthMax: If you wrote the above query then we would assign different variables to the two pattern predicates. So `Found` would still not occure.

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
DarthMax: Yes, we need this for planning of the pattern predicates

caps-core/src/main/scala/org/opencypher/caps/ir/impl/ExpressionConverter.scala
DarthMax: Exactly :)

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
s1ck: I like them as a kind of documentation of intent. Typically, one would disable assertions when compiling for production systems. We already use them in other places.

boggle: Even better: encode the invariant in the type system :)  like `NonEmptyList[...]` ;)

Mats-SX: I guess I'm just not used to this practice.

PR #231, opened 2017-12-06
Add support for pattern expressions via subqueries

s1ck:


caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
```diff
@@ -354,8 +353,8 @@ final case class FalseLit() extends BoolLit(false)()
 
 // Pattern Predicate Expression
 
-final case class PatternExpr(predicateField: Var, pattern: Pattern[Expr])(val cypherType: CypherType = CTBoolean) extends Expr {
+final case class ExistsPatternExpr(predicateField: Var, ir: CypherQuery[Expr])(val cypherType: CypherType = CTBoolean) extends Expr {
```
Mats-SX: Hm, will this embedding of the IR into the expression not cause layering issues for subsequent phases?

Mats-SX: I mean, the IR is consumed by the LogicalPlanner; but now (some of) it is also propagated?

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
```diff
@@ -177,41 +177,53 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
...
         producer.planFilter(v, acc)
-      case (acc, pe: PatternExpr) =>
-        val patternPlan = planComponentPattern(acc, pe.pattern, context.sourceGraph)
-        val withPredicate = producer.planPatternPredicate(pe, acc, patternPlan)
-        producer.planFilter(pe, withPredicate)
+
+      case (acc, ex: ExistsPatternExpr) =>
+        val innerPlan = this(ex.ir)
+        val Predicate = producer.planExistsPatternPredicate(ex, acc, innerPlan)
+        producer.planFilter(ex, Predicate)
```
Mats-SX: Should use lower-case initial character, otherwise this looks like an object.

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
Mats-SX: OTOH, it does heavily signal the subquery aspects of this feature, which is appropriate.

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
```diff
@@ -222,23 +234,33 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
...
       case func: FunctionExpr =>
         val projectArg = planInnerExpr(func.expr, in)
         producer.projectExpr(func, projectArg)
-      case pe : PatternExpr =>
-        val patternPlan = planComponentPattern(in, pe.pattern, context.sourceGraph)
-        producer.planPatternPredicate(pe, in, patternPlan)
+
+      case ex: ExistsPatternExpr =>
+        val innerPlan = this(ex.ir)
+        producer.planExistsPatternPredicate(ex, in, innerPlan)
```
Mats-SX: Should we allow pattern predicates as inner expressions?

Mats-SX: Ah, for `AND`, `OR` etc of course, although this permits it for any expression position.

caps-core/src/main/scala/org/opencypher/caps/impl/parse/rewriter/extractSubqueryFromPatternExpression.scala
```diff
@@ -0,0 +1,98 @@
...
+    /**
+      * WHERE (a)-[:R]->({foo:true})-->()... AND a.age > 20
+      *
+      * to
+      *
+      * WHERE EXISTS (
+      *   MATCH (a)-[e0]->(v0)-[e1]->(v1)...
+      *   WHERE e0:R AND v0.foo = true
+      *   RETURN a, true
+      * ) AND a.age > 20
```
Mats-SX: I think the parentheses should be `{` and `}` here.

caps-core/src/main/scala/org/opencypher/caps/impl/util/MapUtils.scala
```diff
@@ -25,6 +25,18 @@ trait MapUtils {
         case (k, Some(v1), _) => k -> v1
         case (k, _, Some(v2)) => k -> v2
       }.toMap
+
+  def submap[A, B](keys: Seq[A], a: Map[A, B]): Map[A, B] = {
+    val filteredKeys = keys.filter(a.contains)
+    filteredKeys.zip(filteredKeys.map(a)).toMap
```
Mats-SX: Isn't this just `a.filterKeys(keys.contains)`?

caps-core/src/test/scala/org/opencypher/caps/impl/typer/SchemaTyperTest.scala
```diff
@@ -325,10 +325,6 @@ class SchemaTyperTest extends BaseTestSuite with Neo4jAstTestSupport with Mockit
     // assertExpr.from("percentileDisc([1, 3.14][$param], 3.14)") shouldHaveInferredType CTInteger
   }
 
-  test("typing of pattern expression") {
-    assertExpr.from("(a)-[:KNOWS]->(b)") shouldHaveInferredType CTBoolean
-  }
```
Mats-SX: Why was this deleted?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/SparkSQLExprMapper.scala
```diff
@@ -71,6 +71,7 @@ object SparkSQLExprMapper {
       * @param expr    the expression to be evaluated.
       * @param df      the dataframe containing the data over which the expression should be evaluated.
       * @param context context with helper functions, such as column names.
+      * @param context context with helper functions, such as column names.
```
Mats-SX: Duplicate doc row

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
```diff
@@ -53,10 +53,20 @@ final case class ValueJoin(
   }
 
 }
-abstract class PatternJoin extends BinaryPhysicalOperator {
+
+/**
+  * This operator performs a left outer join between the already matched path and the optional matched pattern and
+  * updates the resulting columns.
+  *
+  * @param lhs mandatory match data
```
Mats-SX: let's not use `mandatory`; I prefer `non-optional` or `matched` or `definitely matched`

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
```diff
@@ -53,10 +53,20 @@ final case class ValueJoin(
...
+  * @param rhs optional match data
+  * @param header result header (lhs header + rhs header)
+  */
+final case class Optional(lhs: PhysicalOperator, rhs: PhysicalOperator, header: RecordHeader)
+    extends BinaryPhysicalOperator {
 
   override def executeBinary(left: PhysicalResult, right: PhysicalResult)
-                            (implicit context: RuntimeContext): PhysicalResult = {
+      (implicit context: RuntimeContext): PhysicalResult = {
```
Mats-SX: strange indent change

caps-core/src/main/scala/org/opencypher/caps/api/expr/Expr.scala
DarthMax: Yes, so we only do this for this single expression which clearly signals: This is a subquery, do plan it.

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
DarthMax: It does, but the semantic check do not allow it I guess

caps-core/src/test/scala/org/opencypher/caps/impl/typer/SchemaTyperTest.scala
DarthMax: because the test helper does not call rewriters and sub queries can not yet be parsed. We building the expression tree by hand but it was a mess and would not have contributed much in terms of stability etc

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
Mats-SX: It does, but that's another debate really. (Because in Cypher 3.x pattern expressions evaluate to a list of paths and can be used in any expression context.)

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/BinaryOperators.scala
DarthMax: mhh I don't find a better way tbh

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
DarthMax: mhh I guess that is a point. On the other hand we do not really have a good way to restrict this in the planInnerExpression method right now. We could introduce special cases for `NOT`, `ANDS` and `ORS` but that does not feel good

caps-core/src/main/scala/org/opencypher/caps/impl/util/MapUtils.scala
s1ck: You're right. We weren't away of that.

s1ck: However, I think we should keep the method as it's much more readable.

caps-core/src/main/scala/org/opencypher/caps/impl/logical/LogicalPlanner.scala
Mats-SX: I agree.

PR #236, opened 2017-12-21
Move remaining efficient implementations to TreeNode

pstutz:


caps-core/src/main/scala/org/opencypher/caps/impl/common/TreeNode.scala
```diff
@@ -18,8 +18,16 @@ package org.opencypher.caps.impl.common
 import scala.reflect.ClassTag
 import scala.util.hashing.MurmurHash3
 
+/**
+  * This is the basic tree node class. Usually it makes more sense to use `AbstractTreeNode`, which uses reflection
+  * to generate the `children` and `withNewChildren` field/method. Our benchmarks show that manually implementing
+  * them can result in a speedup of a factor of ~3 for tree rewrites, so in performance critical places it might
+  * make sense to extend `TreeNode` instead.
```
s1ck: good doc!

PR #241, opened 2018-01-10
Update frontend dependency to 3.4.0-alpha05

Mats-SX:
This commit only guarantees successful compilation


caps-core/src/main/scala/org/opencypher/caps/impl/parse/CypherParser.scala
```diff
@@ -15,14 +15,14 @@
...
+import org.neo4j.cypher.internal.frontend.v3_4.ast.rewriters._
+import org.neo4j.cypher.internal.frontend.v3_4.helpers.rewriting.RewriterStepSequencer
+import org.neo4j.cypher.internal.frontend.v3_4.phases._
+import org.neo4j.cypher.internal.frontend.v3_4.semantics._
 import org.opencypher.caps.impl.CompilationStage
-import org.opencypher.caps.impl.parse.rewriter.CAPSRewriting
 import org.opencypher.caps.impl.exception.Raise
+import org.opencypher.caps.impl.parse.rewriter.CAPSRewriting
 
 object CypherParser extends CypherParser {
```
s1ck: in https://github.com/neo4j/neo4j/blob/3.4/community/cypher/frontend-3.4/src/main/scala/org/neo4j/cypher/internal/frontend/v3_4/phases/AstRewriting.scala#L26 we also need to specify the boolean parameter for disabling getDegree Rewriting (similar to what's happening in extractSubqueryFromPatternExpression.scala)

Mats-SX: It still worked though... but I guess that's because we didn't write tests for these cases.

PR #242, opened 2018-01-11
Restructuring of Test environment

DarthMax:
This PR does a complete refactoring of the tests environment:

- Introduce parser for Cypher CREATE and UNWIND statements
- use in memory PropertyGraph representation as intermediate test graph
- infrastructure to convert property graph into different CAPSGraph representations
- use ScalaTest for TCK implementation


caps-core/src/test/scala/org/opencypher/caps/test/BaseTestSuite.scala
```diff
@@ -18,12 +18,20 @@ package org.opencypher.caps.test
...
+  extends FunSpec
   with Matchers
   with AllInstances
-  with AllSyntax
+  with AllSyntax {
+
+  /**
+    * Wraps an 'it' call for convenience reasons
+    */
+  def test(name: String, tags: Tag*)(testFun: => Any /* Assertion */)(implicit pos: source.Position): Unit =
```
Mats-SX: Yay :)

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptanceFunSpecMixin/AggregationBehaviour.scala
```diff
@@ -0,0 +1,630 @@
...
+import org.opencypher.caps.api.value.CypherMap
+
+import scala.collection.immutable.Bag
+
+trait AggregationBehaviour {
+  this: AcceptanceTest =>
+
+  def aggregationBehaviour(initGraph: String => CAPSGraph): Unit = {
+
+    describe("AVG") {
```
Mats-SX: I'm a fan of the grouping feature already. Makes it easier to observe coverage.

I'm just thinking that many of these tests would do great in the TCK. I had the idea that we'd convert these tests to Gherkin format and use the TCK tools to run them like that, which would make it even easier to move them over to the TCK long-term... but this is still great of course.

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptanceFunSpecMixin/FunctionsBehaviour.scala
```diff
@@ -0,0 +1,254 @@
...
+      keysAsMap should equal(Bag(
+        CypherMap("k" -> CypherList(Seq("age", "eyes", "name")))
+      ))
+    }
+
+    test("keys() does not return keys of unset properties") {
+      val given = initGraph(
+        """
+          |CREATE (:Person {name:'Alice', age: 64, eyes:'brown'})
+          |CREATE (:Person {name:'Bob', eyes:'blue'})
```
Mats-SX: The fact that this is now real Cypher makes me happy :)

caps-spark/src/test/scala/org/opencypher/caps/test/support/DebugOutputSupport.scala
```diff
@@ -48,6 +48,11 @@ trait DebugOutputSupport {
     def toBag: Bag[Row] = Bag(elements.toSeq: _*)
   }
 
+  implicit class GenericIterableToBagConverter[T](val elements: Iterable[T]) {
```
Mats-SX: Could even be `TraversableOnce[T]` to cover more types (not necessary in this PR)

caps-spark/src/test/scala/org/opencypher/caps/test/support/creation/caps/CAPSScanGraphFactoryTest.scala
```diff
@@ -0,0 +1,91 @@
...
+    """
+      |CREATE (max:Person:Astronaut {name: "Max"})
+      |CREATE (martin:Person:Martian {name: "Martin"})
+      |CREATE (swedish:Language {title: "Swedish"})
+      |CREATE (german:Language {title: "German"})
+      |CREATE (orbital:Language {title: "Orbital"})
+      |CREATE (max)-[:SPEAKS]->(swedish)
+      |CREATE (max)-[:SPEAKS]->(german)
+      |CREATE (martin)-[:SPEAKS]->(german)
+      |CREATE (martin)-[:SPEAKS]->(orbital)
```
Mats-SX: 👽

caps-spark/src/test/scala/org/opencypher/caps/test/support/creation/propertygraph/Neo4jPropertyGraphFactory.scala
```diff
@@ -0,0 +1,67 @@
...
+      val relType: String = neoRel.getType.name
+      val sourceId: Long = neoRel.getStartNodeId
+      val targetId: Long = neoRel.getEndNodeId
+      val id: Long = neoRel.getId
+      val properties: Map[String, Any] = neoRel.getAllProperties.asScala.toMap
+
+      Relationship(id, sourceId, targetId, relType, properties)
+    }.toSeq
+
+    tx.success()
```
Mats-SX: Don't we also need `tx.close()` here?

caps-tck/src/test/scala/org/opencypher/caps/TCKCAPSTest.scala
```diff
@@ -0,0 +1,93 @@
...
+    "scenario",
+    scenarios.filterNot { s => ScenarioBlacklist.contains(s.toString()) }: _*
+  )
+
+  val blackListScenarios = Table(
+    "scenario",
+    scenarios.filter { s => ScenarioBlacklist.contains(s.toString()) }.groupBy(_.toString()).map(_._2.head).toSeq: _*
+  )
+
+  // white list tests are run on all factories
```
Mats-SX: So there is one global blacklist for all factories then? One could imagine that some scenario works on one factory but not others -- that wouldn't be supported by this model if I understand it correctly. 

Now, perhaps we should simply not accept that situation; since most logic lives in the engine and the factories really only implement a currently very small interface (`.nodes` and `.rels`) we should be able to keep these in lock-step of one another in terms of capabilities.

DarthMax: For now that is correct. I propose we leave it as it is until we have more CAPS graph factories and then see what happens. But I agree that this situation should not occur after all.

PR #248, opened 2018-01-22
Extract minimal part of CAPSSession to public API

Mats-SX:


caps-spark/src/main/scala/org/opencypher/caps/api/CAPSSession.scala
```diff
@@ -30,9 +30,10 @@ import org.opencypher.caps.impl.spark.io.hdfs.HdfsCsvPropertyGraphDataSourceFact
 import org.opencypher.caps.impl.spark.io.neo4j.Neo4JPropertyGraphDataSourceFactory
 import org.opencypher.caps.impl.spark.io.session.SessionPropertyGraphDataSourceFactory
 
-trait CAPSSession extends CypherSession {
+trait CAPSSession extends CypherSession with CAPSSessionOps {
```
Mats-SX: Doesn't need to be on this trait since it's on the impl trait. I'm unsure whether we want to offer this operation-based API in version 1.

caps-spark/src/main/scala/org/opencypher/caps/api/spark/CAPSSessionOps.scala
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.opencypher.caps.api.spark
+
+import org.opencypher.caps.api.graph.PropertyGraph
+import org.opencypher.caps.api.record.CypherRecords
+import org.opencypher.caps.api.value.CypherValue
+import org.opencypher.caps.ir.api.expr.{Expr, Var}
+
+trait CAPSSessionOps {
```
Mats-SX: Nice!

pstutz: Thanks, I just removed session ops from the session trait.

PR #251, opened 2018-01-23
Service discovery for `CAPSPropertyGraphSourceFactory`

pstutz:


caps-api/src/main/scala/org/opencypher/caps/impl/util/MapUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.opencypher.caps.impl.util
+
+// TODO: Replace with Cats Monoid
+trait MapUtils {
```
Mats-SX: Why is this file back in? It was removed in the previous PR

caps-api/src/test/scala/org/opencypher/caps/impl/util/MapUtilsTest.scala
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.opencypher.caps.impl.util
+
+import org.opencypher.caps.test.BaseTestSuite
+
+class MapUtilsTest extends BaseTestSuite {
```
Mats-SX: Same as above; should be gone.

caps-api/src/main/scala/org/opencypher/caps/impl/util/MapUtils.scala
pstutz: Thanks, good find, it probably slipped in during rebase.

PR #252, opened 2018-01-24
Add support for List type to CsvGraphLoader

DarthMax:


caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvGraphLoader.scala
```diff
@@ -116,12 +117,14 @@ class CsvGraphLoader(fileHandler: CsvGraphLoaderFileHandler)(implicit capsSessio
...
       val schema = parseSchema(e)(CsvNodeSchema(_))
 
-      val records = CAPSRecords.create(
-        sparkSession.read
-          .option("timestampFormat", "yyyy-MM-dd'T'HH:mm:ss.SSS")
-          .schema(schema.toStructType)
-          .csv(e.toString)
-      )
+      val dataFrame = sparkSession.read
+        .option("timestampFormat", "yyyy-MM-dd'T'HH:mm:ss.SSS")
```
s1ck: Have you tried running it without the config option? I think this was a workaround for a bug before Spark 2.2.0

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvGraphLoader.scala
```diff
@@ -174,6 +181,23 @@ class CsvGraphLoader(fileHandler: CsvGraphLoaderFileHandler)(implicit capsSessio
...
+
+  /**
+    * CSV does not allow to read array fields. Thus we have to read all list fields as Strings and then manually convert
+    * them
+    *
+    * @param dataFrame the input data frame with list column represented as string
+    * @param schema the dataframes csv schema
+    * @return Date frame with array list columns
+    */
+  private def convertLists(dataFrame: DataFrame, schema: CsvSchema): DataFrame = {
```
s1ck: Nice

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvSchema.scala
```diff
@@ -28,15 +28,42 @@ abstract class CsvSchema {
 }
 
 case class CsvField(name: String, column: Int, valueType: String) {
-  def getType: DataType = valueType.toLowerCase match {
-    case "string"  => StringType
-    case "integer" => LongType
-    case "long"    => LongType
-    case "boolean" => BooleanType
-    case x         => throw new RuntimeException(s"Unknown type $x")
+  private val listType = raw"list\[(\w+)\]".r
```
s1ck: maybe instantiate the pattern here once because you only use the pattern later.
edit: you need also the raw string, but then it's maybe better to instatiate the pattern to a separate val as this could be called a lot during import

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvSchema.scala
```diff
@@ -28,15 +28,42 @@ abstract class CsvSchema {
...
+    case other => extractSimpleType(other)
+  }
+
+  /**
+    * For List types we return the target array type.
+    * @return the Spark SQL type of the csv field at after special conversions
+    */
+  def getTargetType: DataType = valueType.toLowerCase match {
+    case l if listType.pattern.matcher(l).matches() => l match {
+      case listType(inner) => ArrayType(extractSimpleType(inner))
```
s1ck: ah, that is cool. ok, then you need the raw string separately

caps-spark/src/test/resources/csv/sn/nodes/people.csv
```diff
@@ -1,4 +1,4 @@
-1,Stefan,42,false,true
-2,Mats,23,true,false,
-3,Martin,1337,false,true
-4,Max,8,false,true
\ No newline at end of file
+1,Stefan,42,false,true,"german|english"
```
s1ck: Maybe we should make these separators configurable at some point, i.e. token, line and array separators. I hope the spark csv parsing is closer to standardized parsing than Flink is :)

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvSchema.scala
DarthMax: That is already the pattern, the little `.r` at the end converts it into a Regex. 
I have converted all functions into lazy vals though.

caps-spark/src/test/resources/csv/sn/nodes/people.csv
DarthMax: Yes I had the same thought. I'm not sure if it should be a part of the schema or of the graph loader

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvGraphLoader.scala
DarthMax: seems to work without :)

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/CAPSSessionFileTest.scala
```diff
@@ -49,10 +49,10 @@ class CAPSSessionFileTest extends CAPSTestSuite {
...
     */
   def testGraphNodes: Bag[Row] = Bag(
-    Row(1L, true, true, true, false, 42L, "Stefan"),
-    Row(2L, true, false, true, true, 23L, "Mats"),
-    Row(3L, true, true, true, false, 1337L, "Martin"),
-    Row(4L, true, true, true, false, 8L, "Max")
+    Row(1L, true,  true, true, false, mutable.WrappedArray.make(Array("german", "english")),            42L, "Stefan"),
+    Row(2L, true, false, true,  true, mutable.WrappedArray.make(Array("swedish", "english", "german")), 23L,   "Mats"),
+    Row(3L, true,  true, true, false, mutable.WrappedArray.make(Array("german", "english")),            1337L, "Martin"),
+    Row(4L, true,  true, true, false, mutable.WrappedArray.make(Array("german", "swedish", "english")), 8L,    "Max")
```
Mats-SX: Does it need to be a wrapped array? Seems like a harsh restriction. I suspect that you did it like this because using only an `Array` fails at some point in the runtime?

PR #253, opened 2018-01-24
Refactor CypherSession API and examples

s1ck:


s1ck:
@Mats-SX @DarthMax maybe start with `MultipleGraphExample.scala` to see the relevant API changes.

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
```diff
@@ -18,73 +18,82 @@ package org.opencypher.caps.api.graph
...
 import org.opencypher.caps.api.io.{CreateOrFail, PersistMode, PropertyGraphDataSource}
+import org.opencypher.caps.api.schema.Schema
 import org.opencypher.caps.api.value.CypherValue
+import org.opencypher.caps.impl.record.CypherRecords
 
+// TODO: extend doc with explanation for writing graphs
 /**
-  * The session that manages graphs and adds Cypher query capabilities to them.
-  * Sessions serve as containers for configuration and provide graph loading via URIs.
+  * The Cypher Session is the main entry point for a CAPS application. It manages graphs which can be queried using
```
DarthMax: Maybe we should not call it `CAPS` application

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
```diff
@@ -18,73 +18,82 @@ package org.opencypher.caps.api.graph
...
+  private[graph] def cypherOnGraph(graph: PropertyGraph, query: String, parameters: Map[String, CypherValue] = Map.empty): CypherResult
+
+  /**
+    * Reads a graph from the argument URI.
+    *
+    * @param uri URI locating a graph
+    * @return graph located at the URI
     */
-  def cypher(graph: PropertyGraph, query: String, parameters: Map[String, CypherValue]): CypherResult
+  def readFromURI(uri: URI): PropertyGraph
```
DarthMax: To me it feels that readFrom is a bit redundant

caps-spark/src/main/scala/org/opencypher/caps/api/CAPSSession.scala
```diff
@@ -20,17 +20,30 @@ import java.util.{ServiceLoader, UUID}
...
 import scala.collection.JavaConverters._
+import scala.reflect.runtime.universe._
 
 trait CAPSSession extends CypherSession {
 
   def sparkSession: SparkSession
 
+  def readFromSeqs[N <: Node : TypeTag, R <: Relationship : TypeTag](
+    nodes: Seq[N],
+    relationships: Seq[R]): PropertyGraph = {
```
DarthMax: I think relationships could be optional

s1ck:
@DarthMax the methods you mentioned are for creating dataframes from some specific input (e.g. scala / java collection). We were looking into the methods for reading from different formats. In Spark this is done via: 
`spark.read.format("json").load("examples/src/main/resources/people.json")`
where 'json' could be replaced with 'csv', 'neo4j' ...

s1ck:
I personally like those fluent APIs much more than the current URI scheme. Less error-prone and writable without looking into docs.

DarthMax:
It is somewhat of a mix, isn't it? Especially the readFromSeqs, which is closer to `SparkSession.createDataFrame(Seq[Product])`. 

I like the idea of a fluent API for the specialized data sources like neo4j, csv, etc

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherResult.scala
```diff
@@ -27,11 +27,12 @@ import org.opencypher.caps.trees.TreeNode
 trait CypherResult extends CypherPrintable {
 
   /**
-    * Retrieves the single graph returned by the query, if it returned exactly one graph.
+    * Retrieves the graph returned by the query. Note, if the query returns more than one graph, there is no guarantee
+    * about which graph is returned.
     *
-    * @return the single graph, otherwise None.
+    * @return a graph
```
Mats-SX: One of the graphs returned by the query

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
```diff
@@ -20,71 +20,69 @@ import java.net.URI
...
-    * @param query      the Cypher query to execute.
-    * @param parameters the parameters used by the Cypher query.
-    * @return the result of the query.
+    * @param graph      ambient graph for this query
+    * @param query      Cypher query to execute
+    * @param parameters parameters used by the Cypher query
+    * @return result of the query
     */
-  def cypher(graph: PropertyGraph, query: String, parameters: Map[String, CypherValue]): CypherResult
+  private[graph] def cypherOnGraph(graph: PropertyGraph, query: String, parameters: Map[String, CypherValue] = Map.empty): CypherResult
```
Mats-SX: Is this hidden because the canonical way is to do `graph.cypher(query)` instead of `session.cypher(graph, query)

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
```diff
@@ -20,71 +20,69 @@ import java.net.URI
...
-    * @param graph the graph to store.
-    * @param uri the graph URI indicating location and format to store the graph in.
-    * @param mode the persist mode which determines what happens if the location is occupied.
-    * @return the stored graph.
+    * @param graph graph to write
+    * @param uri   graph URI indicating location and format to write the graph to
+    * @param mode  persist mode which determines what happens if the location is occupied
     */
-  def storeGraphAt(graph: PropertyGraph, uri: String, mode: PersistMode = CreateOrFail): PropertyGraph
+  // TODO: Error handling via Return Type (Success, Failed) or just throw exception?
```
Mats-SX: Since we're returning `Unit` here anyway, we might as well indicate success/failure on the return value IMO -- but perhaps it's not acute enough?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/CAPSSessionImpl.scala
```diff
@@ -80,9 +80,10 @@ sealed class CAPSSessionImpl(val sparkSession: SparkSession, private val graphSo
   def unmountAll(): Unit =
     graphSourceHandler.unmountAll(this)
 
-  override def emptyGraph: CAPSGraph = CAPSGraph.empty(this)
+  override def cypher(query: String, parameters: Map[String, CypherValue]): CypherResult =
+    cypherOnGraph(CAPSGraph.empty(this), query, parameters)
 
-  override def cypher(graph: PropertyGraph, query: String, queryParameters: Map[String, CypherValue]): CypherResult = {
+  override def cypherOnGraph(graph: PropertyGraph, query: String, queryParameters: Map[String, CypherValue]): CypherResult = {
```
Mats-SX: How does Scala access limits work through inheritance? E.g. do we need to restate `private[spark]` here or something?

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
pstutz: Yes, we think that the API is easier to discover and learn if there is only one canonical way to do the same thing.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/CAPSSessionImpl.scala
pstutz: I'm not sure what the conventions are. My thinking is that an API user should never encounter the concrete implementation of the type, but if someone does, then we assume that they know what they are doing.

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherResult.scala
```diff
@@ -27,11 +27,12 @@ import org.opencypher.caps.trees.TreeNode
 trait CypherResult extends CypherPrintable {
 
   /**
-    * Retrieves the single graph returned by the query, if it returned exactly one graph.
+    * Retrieves the graph returned by the query. Note, if the query returns more than one graph, there is no guarantee
```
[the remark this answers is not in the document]
pstutz: We wanted to be explicit that we don't guarantee which one we return. I guess a user could also just assume that.

caps-spark/src/test/scala/org/opencypher/caps/demo/MultiGraphExample.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  // 7) Query for product recommendations
+  val recommendations = recommendationGraph.cypher(
+    """|MATCH (person:Person)-[:FRIEND_OF]-(friend:Person),
+       |(friend)-[:IS]->(customer:Customer),
+       |(customer)-[:BOUGHT]->(product:Product)
+       |RETURN DISTINCT product.title AS recommendation, person.name AS for
+    """.stripMargin)
+
+  recommendations.print
```
Mats-SX: Love the example!

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherSession.scala
pstutz: We also aren't sure what to do here, so for now, we added the TODO item.

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherResult.scala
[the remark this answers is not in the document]
Mats-SX: I meant only the first sentence. The second sentence is good.

PR #258, opened 2018-01-26
Separate API Cypher value trait from CAPS Cypher value implementation

pstutz:


caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalPlanner.scala
```diff
@@ -139,16 +150,16 @@ class PhysicalPlanner extends DirectCompilationStage[FlatOperator, PhysicalOpera
...
+      rel,
+      edgeList,
+      target,
+      lower,
+      upper,
+      sourceOp,
+      relOp,
+      targetOp,
+      header,
+      isExpandInto) =>
```
Mats-SX: This indentation also looks weird.

caps-tck/src/test/scala/org/opencypher/caps/TCKFixture.scala
```diff
@@ -19,7 +19,7 @@ import org.opencypher.caps.api.CAPSSession
 import org.opencypher.caps.api.exception.NotImplementedException
 import org.opencypher.caps.impl.record.CypherRecords
 import org.opencypher.caps.impl.spark.CAPSGraph
-import org.opencypher.caps.api.value.{CypherValue => CAPSValue}
+import org.opencypher.caps.api.value.{CAPSValue => CAPSValue}
```
Mats-SX: Could be removed now

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalPlanner.scala
pstutz: Our IntelliJ auto-formatter likes it that way. :(

caps-api/src/main/scala/org/opencypher/caps/api/value/syntax/CypherValueSyntax.scala
```diff
@@ -1,92 +0,0 @@
...
-    new CypherValueOps[V](value)
-}
-
-final class CypherMapOps[V <: CypherMap](val value: V) extends AnyVal with Serializable {
-  def properties(implicit companion: CypherMapCompanion[V]): Option[Properties] =
-    companion.properties(value)
-}
-
-final class CypherEntityOps[V <: CypherEntityValue](val value: V) extends AnyVal with Serializable {
-  def id(implicit companion: CypherEntityCompanion[V]): Option[EntityId] =
```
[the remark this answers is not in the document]
pstutz: Neo4J and CAPS use Long IDs internally. Does the specification require this as well? If so, then I think it makes sense to give access to them. If not, then I think it might be better not to expose the ID.

I will add CypherNode and CypherRelationship and expose labels/relType and properties in the trait itself.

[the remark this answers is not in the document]
Mats-SX: No, the id is not specified to be a number, only a unique value for all nodes and relationships respectively (could be a string f.e.), so we don't need to expose it I think.

PR #260, opened 2018-01-30
Add support for undirected patterns

Mats-SX:
Undirected pattern matches are implemented by taking the union of expanding in each direction.


caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/ReturnBehaviour.scala
```diff
@@ -270,16 +270,31 @@ trait ReturnBehaviour { this: AcceptanceTest =>
...
+      }
+
+      it("can evaluate limit with parameter value"){
+        val graph = initGraph("CREATE (a:A),(b:B),(c:C)")
+
+        val res = graph.cypher(
+          """
+            |MATCH (a)
+            |WITH a
+            |LIMIT {limit}
```
Mats-SX: This parameter syntax is deprecated. It should be `$limit`. Did you find this in the TCK?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/TernaryOperators.scala
```diff
@@ -177,15 +176,8 @@ final case class BoundedVarExpand(
...
-          case ProjectedExpr(_ : StartNode) => true
-          case _ => false
-        }.get)
-        val endNodeSlot = columnName(secondRecords.header.slots.map(_.content).find {
-          case ProjectedExpr(_ : EndNode) => true
-          case _ => false
-        }.get)
-
+        val startNodeSlot = columnName(secondRecords.header.sourceNodeSlot(rel))
+        val endNodeSlot = columnName(secondRecords.header.targetNodeSlot(rel))
```
Mats-SX: Nice improvement.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalPlanner.scala
```diff
@@ -137,29 +138,37 @@ class PhysicalPlanner extends DirectCompilationStage[FlatOperator, PhysicalOpera
...
-
-      case op@flat.ExpandInto(source, rel, target, sourceOp, header, relHeader) =>
-        val in = process(sourceOp)
+        direction match {
+          case Directed =>
+            ExpandSource(first, second, third, source, rel, target, header)
+          case Undirected =>
+            val outgoing = ExpandSource(first, second, third, source, rel, target, header)
+            val incoming = ExpandSource(third, second, first, target, rel, source, header, removeSelfRelationships = true)
+            Union(outgoing, incoming)
```
Mats-SX: I realised that an alternative to this implementation might be to do a join with a disjunctive predicate; instead of joining on `a = source(r)` and then `b = end(r)` we could use `a = source(r) OR a = end(r)` and then `b = end(r) OR b = source(r)`.

That would then result in two joins per expand instead of four joins plus union for undirected.

DarthMax: This is a perfect case for benchmark backed performance improvement :)

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/ReturnBehaviour.scala
DarthMax: From Alex LDBC queries

PR #269, opened 2018-02-05
Refactor CypherValue

pstutz:


caps-api/src/main/scala/org/opencypher/caps/api/value/CypherValue.scala
```diff
@@ -15,47 +15,345 @@
...
+      */
+    override def equals(other: Any): Boolean = {
+      other match {
+        case cv: CypherValue => Objects.equals(unwrap, cv.unwrap)
+        case _ => false
+      }
+    }
+
+    /**
+      * A Cypher string representation.
```
Mats-SX: We could refer to the openCypher TCK for the chosen format of these strings. Perhaps something like this:

> For more information about the exact format of these, please refer to the openCypher TCK 

And a link to https://github.com/opencypher/openCypher/tree/master/tck#format-of-the-expected-results

caps-api/src/main/scala/org/opencypher/caps/api/value/CypherValue.scala
```diff
@@ -15,47 +15,345 @@
...
+    override def unwrap: List[Any] = value.map(_.unwrap)
+  }
+
+  object CypherList extends UnapplyValue[List[CypherValue], CypherList] {
+    def apply(elem: Any*): CypherList = elem.map(CypherValue(_)).toList
+
+    val empty: CypherList = List.empty[CypherValue]
+  }
+
+  sealed trait CypherEntity[+Id] extends Product with MaterialCypherValue[CypherEntity[Id]] {
```
Mats-SX: What is the advantage about extending `Product` for the entities? Is is to enable easier inheritance by a `case class`?

caps-ir/src/main/scala/org/opencypher/caps/ir/impl/typer/SchemaTyper.scala
```diff
@@ -349,15 +333,21 @@ object SchemaTyper {
       expr: Expression,
       orderedExprs: Vector[Expression]): Eff[R, CypherType] = {
     for {
-      innerTypes <- EffMonad[R].sequence(orderedExprs.map(process[R]))
+      hasLabelPreds <- EffMonad[R].sequence(orderedExprs.collect { case h: HasLabels => h }.map(process[R]))
```
Mats-SX: Since we forgot about it from last time, perhaps we should add a comment here detailing why we need to process `HasLabels` first?

```
/** 
  * We have to process label predicates first in order to use label information on node
  * variables for looking up the type of property expressions on these variables.
  */
```

caps-spark/src/main/scala/org/opencypher/caps/api/value/CAPSEntity.scala
```diff
@@ -15,19 +15,16 @@
...
-  self =>
-
-  override def toString = s"#$v"
-}
+case class CAPSRelationship(
+  override val id: Long,
+  override val source: Long,
+  override val target: Long,
+  override val relType: String,
+  override val properties: CypherMap = CypherMap.empty) extends CypherRelationship(id, source, target, relType, properties)
```
Mats-SX: Very succinct and clear!

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/DfUtils.scala
```diff
@@ -18,62 +18,26 @@ package org.opencypher.caps.impl.spark
...
-      // TODO: This supports var-expand where we only track rel ids, but it's not right
-      case _: CTRelationship =>
-        (in) =>
-          cypherInteger(in.asInstanceOf[Long])
-      case l: CTList =>
-        (in) =>
-          {
-            val converted = in.asInstanceOf[Seq[_]].map(typeToValue(l.elementType))
-            cypherList(converted.toIndexedSeq)
+              CypherValue(r.get(index))
```
Mats-SX: Wow! This was a huge win!

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/SparkSQLExprMapper.scala
```diff
@@ -44,9 +44,12 @@ object SparkSQLExprMapper {
     implicit context: RuntimeContext): Column = {
     expr match {
       case p@Param(name) if p.cypherType.subTypeOf(CTList(CTAny)).maybeTrue =>
-        udf(const(context.parameters(name)), toSparkType(p.cypherType))()
+        context.parameters(name) match {
+          case CypherList(l) => functions.array(l.unwrap.map(functions.lit(_)): _*)
```
Mats-SX: Avoiding a UDF: nice!

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/SparkSQLExprMapper.scala
```diff
@@ -160,10 +163,29 @@ object SparkSQLExprMapper {
...
 
-        case inEq: LessThan => Some(inequality(lt, header, inEq, df))
-        case inEq: LessThanOrEqual => Some(inequality(lteq, header, inEq, df))
-        case inEq: GreaterThanOrEqual => Some(inequality(gteq, header, inEq, df))
-        case inEq: GreaterThan => Some(inequality(gt, header, inEq, df))
+        case LessThan(lhs, rhs) =>
+          val l = getColumn(lhs, header, df)
+          val r = getColumn(rhs, header, df)
+
+          Some(l < r)
```
Mats-SX: Reader's note: this is possible without violating Cypher semantics because

1) Spark SQL gives `null` back when comparing across types (from initial investigation)
2) We never have multiple types per column in CAPS (yet)

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/Udfs.scala
```diff
@@ -16,26 +16,13 @@
...
-
-  // TODO: Try to share code with cypherFilter()
-  def lt(lhs: Any, rhs: Any): Any = (CAPSValue(lhs) < CAPSValue(rhs)).orNull
-
-  def lteq(lhs: Any, rhs: Any): Any = (CAPSValue(lhs) <= CAPSValue(rhs)).orNull
-
-  def gteq(lhs: Any, rhs: Any): Any = (CAPSValue(lhs) >= CAPSValue(rhs)).orNull
-
-  def gt(lhs: Any, rhs: Any): Any = (CAPSValue(lhs) > CAPSValue(rhs)).orNull
+  def const(v: Any): () => Any = () => v
```
Mats-SX: It might even be possible to remove `const` now that we use `array` and `lit` in `getColumn()` above.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/PhysicalPlanner.scala
```diff
@@ -137,7 +136,7 @@ class PhysicalPlanner extends DirectCompilationStage[FlatOperator, PhysicalOpera
 
         direction match {
           case Directed =>
-            ExpandSource(first, second, third, source, rel, target, header)
+        ExpandSource(first, second, third, source, rel, target, header)
```
Mats-SX: This looks weird

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/physical/operators/UnaryOperators.scala
```diff
@@ -91,20 +91,19 @@ final case class Unwind(in: PhysicalOperator, list: Expr, item: Var, header: Rec
...
-              val list = t.map(Row(_)).toList.asJava
-
+          context.parameters(name).as[CypherList] match {
+            case Some(l) =>
               val sparkType = toSparkType(item.cypherType)
               val nullable = item.cypherType.isNullable
               val schema = StructType(Seq(StructField(itemColumn, sparkType, nullable)))
 
-              val df = records.caps.sparkSession.createDataFrame(list, schema)
+              val javaRowList = l.unwrap.map(Row(_)).asJava
```
Mats-SX: Another nice cleanup; no need to match on `TraversableOnce`!

caps-spark/src/test/scala/org/opencypher/caps/api/value/CAPSValueComparabilityTest.scala
```diff
@@ -1,117 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package org.opencypher.caps.api.value
-
-class CAPSValueComparabilityTest extends CAPSValueTestSuite {
```
Mats-SX: This is one I'm a bit sad to see go, but we don't need it right now.

caps-spark/src/test/scala/org/opencypher/caps/api/value/CAPSValueConversionTest.scala
```diff
@@ -15,182 +15,130 @@
...
-      case contents => CAPSRelationship(contents.id, contents.startId, contents.endId, contents.relationshipType, contents.properties)
+      case scalaClass => CypherValue(scalaClass)
     }
-
-    newValues should equal(originalValues)
-
+    assert(newValues == originalValues)
     originalValues.foreach { v =>
-      CAPSRelationship.isOrContainsNull(v) should equal(v == null)
+      assert(v.isNull == (v == CypherNull))
```
Mats-SX: Is this where you said ScalaTest went insane and considered obviously equal values different?

caps-spark/src/test/scala/org/opencypher/caps/api/value/CAPSValueEncodingTest.scala
```diff
@@ -15,87 +15,79 @@
...
-  }
+class CAPSValueEncodingTest extends CAPSValueTestSuite with SparkSessionFixture {
 
   test("RELATIONSHIP encoding") {
     val values = RELATIONSHIP_valueGroups.flatten
-    val ds = session.createDataset[CAPSValue](values)(cypherValueEncoder)
+    val ds = session.createDataset[CypherValue](values)(cypherValueEncoder)
 
-    ds.collect().toSeq should equal(values)
+   ds.collect().toList should equal(values)
```
Mats-SX: Woops, indentation :)

caps-spark/src/test/scala/org/opencypher/caps/api/value/CAPSValueEquivTest.scala
```diff
@@ -1,105 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package org.opencypher.caps.api.value
-
-class CAPSValueEquivTest extends CAPSValueTestSuite {
```
Mats-SX: Recording this one so we may find it one day in the depths of history.

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/CAPSScanGraphTest.scala
```diff
@@ -50,7 +50,7 @@ class CAPSScanGraphTest extends CAPSTestSuite with TeamDataFixture {
         Row(300L, false, true, true, false, "F", 84L, "Eve", null, null),
         Row(400L, false, true, true, false, "R", 49L, "Carl", null, null)
       )
-    )
+          )
```
Mats-SX: Some more weird space changes; maybe due to the git merge option?

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/AggregationBehaviour.scala
```diff
@@ -42,8 +42,8 @@ trait AggregationBehaviour {
 
         val result = graph.cypher("MATCH (n) RETURN AVG(n.val) AS res")
 
-        result.records.toMaps should equal(Bag(
-          CAPSMap("res" -> 4)
+        result.records.iterator.toBag should equal(Bag(
```
Mats-SX: Nice! Perhaps we should add a test helper `.asBag` to shorten this somewhat.

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/CAPSScanGraphAcceptanceTest.scala
```diff
@@ -14,8 +14,49 @@
...
 package org.opencypher.caps.impl.spark.acceptance
+import org.opencypher.caps.api.value.CypherValue._
 import org.opencypher.caps.test.support.creation.caps.{CAPSGraphFactory, CAPSScanGraphFactory}
 
+import scala.collection.Bag
+
 class CAPSScanGraphAcceptanceTest extends AcceptanceTest {
   override def capsGraphFactory: CAPSGraphFactory = CAPSScanGraphFactory
+
+  it("Expand into after var expand") {
```
Mats-SX: I think this test should move to `ExpandIntoBehaviour`

caps-tck/src/test/resources/org/opencypher/caps/scenario_blacklist
```diff
@@ -617,8 +616,6 @@ Feature "AggregationAcceptance": Scenario "Aggregates inside normal functions"
 Feature "AggregationAcceptance": Scenario "Handle aggregates inside non-aggregate expressions"
 Feature "AggregationAcceptance": Scenario "Handle aggregation on functions"
 Feature "AggregationAcceptance": Scenario "Distinct on unbound node"
-Feature "AggregationAcceptance": Scenario "Collect distinct nulls"
-Feature "AggregationAcceptance": Scenario "Collect distinct values mixed with nulls"
```
Mats-SX: I like how every time we do some change, one or two scenarios starts working as a side effect. That's a good pattern.

caps-api/src/main/scala/org/opencypher/caps/api/value/CypherValue.scala
pstutz: Ideally, we could use a case class, `Product` is mainly used to imitate the behaviour of a case class without actually being one. The problem with case classes is that one should avoid extending them. In this case, we want the store implementation to extend CypherNode/CypherRelationship with the concrete Id type, and in spite of this offer pattern matching and equals/hashCode that is consistent with case classes.

Mats-SX: Okay, so the intention is to mimic case class usage for a trait, I get it.

caps-spark/src/test/scala/org/opencypher/caps/api/value/CAPSValueConversionTest.scala
pstutz: Yes, that was this case.

PR #275, opened 2018-02-06
TCK coverage improvements

pstutz:


caps-tck/src/test/scala/org/opencypher/caps/TCKFixture.scala
```diff
@@ -17,33 +17,44 @@ package org.opencypher.caps
 
 import org.opencypher.caps.api.CAPSSession
 import org.opencypher.caps.api.exception.NotImplementedException
-import org.opencypher.caps.api.value.{CypherValue => CAPSCypherValue}
-import CAPSCypherValue.{CypherList => CAPSCypherList, CypherMap => CAPSCypherMap}
+import org.opencypher.caps.api.graph.CypherResult
+import org.opencypher.caps.api.value.CypherValue
+import org.opencypher.caps.api.value.CypherValue.{CypherList => CAPSCypherList, CypherMap => CAPSCypherMap, CypherValue => CAPSCypherValue}
```
s1ck: Why is the renaming necessary?

caps-tck/src/test/scala/org/opencypher/caps/TCKFixture.scala
```diff
@@ -40,10 +45,16 @@ case class TCKGraph(capsGraphFactory: CAPSGraphFactory, graph: CAPSGraph)(implic
...
         // mapValues is lazy, so we force it for debug purposes
-        val capsResult = graph.cypher(query, params.mapValues(tckValueToCAPSValue).view.force)
-        val tckRecords = convertToTckStrings(capsResult.records)
-
-        this -> tckRecords
+        val capsResult = Try(graph.cypher(query, params.mapValues(tckValueToCAPSValue).view.force))
+        capsResult match {
+          case Success(r) => this -> convertToTckStrings(r.records)
+          case Failure(e) =>
+            val phase = TCKErrorPhases.RUNTIME // We have no way to detect errors during compile time yet
```
Mats-SX: If the error is `ParsingException` we know it was compile time, although that is just a subset of compile time errors (we also throw a bunch in `SchemaTyper` etc).

pstutz: Would it make sense to have the same fields as the ones in TCK `ExecutionFailed(errorType: String, phase: String, detail: String)` on `CypherException(msg: String)`? That would also mean renaming `msg` => `detail` and wording them the same way if an error is checked by the TCK. These exceptions are user-facing, so if we should decide to do it, we should probably do it before the release.

Mats-SX: That may be a good idea, but sometimes I think we want to add a bit more detail to the user-facing message than the one-liner that the TCK requires, so perhaps we can have both?

PR #278, opened 2018-02-07
#270 fix cross between match clauses

### LINKED ISSUE #270: Cross Product between multiple match clauses
Planning a cross product between two match clauses does not work, if one of the clauses contains at least one expand. See the following test:

```scala
  it("can evaluate cross Product between multiple match clauses") {
    val graph = initGraph("CREATE (:A {val: 0}), (:B {val: 1})-[:REL]->(:C {val: 2})")

    PrintLogicalPlan.set
    val query =
      """
        |MATCH (a:A)
        |MATCH (b:B)-->(c:C)
        |RETURN a.val, c.val
      """.stripMargin

    graph.cypher(query).records.iterator.toBag should equal(Bag(
      CAPSMap("a.val" -> 0, "c.val" -> 2)
    ))
  }
```

```
|-Select(Vector(a.val :: INTEGER, c.val :: INTEGER), Set(), SolvedQueryModel(Set(a :: NODE, b :: NODE, c :: NODE, a.val :: INTEGER,   UNNAMED25 :: RELATIONSHIP, c.val :: INTEGER),Set(a:A :: BOOLEAN, b:B :: BOOLEAN, c:C :: BOOLEAN, true :: BOOLEAN),Set()))
· |-Project(c.val :: INTEGER, Some(c.val :: INTEGER), SolvedQueryModel(Set(a :: NODE, b :: NODE, c :: NODE, a.val :: INTEGER,   UNNAMED25 :: RELATIONSHIP, c.val :: INTEGER),Set(a:A :: BOOLEAN, b:B :: BOOLEAN, c:C :: BOOLEAN, true :: BOOLEAN),Set()))
· · |-Project(a.val :: INTEGER, Some(a.val :: INTEGER), SolvedQueryModel(Set(a :: NODE, b :: NODE, c :: NODE, a.val :: INTEGER,   UNNAMED25 :: RELATIONSHIP),Set(a:A :: BOOLEAN, b:B :: BOOLEAN, c:C :: BOOLEAN, true :: BOOLEAN),Set()))
· · · |-Filter(true :: BOOLEAN, SolvedQueryModel(Set(a :: NODE, b :: NODE, c :: NODE,   UNNAMED25 :: RELATIONSHIP),Set(a:A :: BOOLEAN, b:B :: BOOLEAN, c:C :: BOOLEAN, true :: BOOLEAN),Set()))
· · · · |-Expand(b :: NODE,   UNNAMED25 :: RELATIONSHIP, c :: NODE, Directed, SolvedQueryModel(Set(a :: NODE, b :: NODE, c :: NODE,   UNNAMED25 :: RELATIONSHIP),Set(a:A :: BOOLEAN),Set()))
· · · · · |-NodeScan(b :: :B NODE, SolvedQueryModel(Set(a :: NODE),Set(a:A :: BOOLEAN, b:B :: BOOLEAN),Set()))
· · · · · · |-SetSourceGraph(LogicalExternalGraph(c149c1b1-3887-4707-9ac8-b9d51f3519dd)(session:///graphs/ambient/c149c1b1-3887-4707-9ac8-b9d51f3519dd), SolvedQueryModel(Set(a :: NODE),Set(a:A :: BOOLEAN),Set()))
· · · · · · · |-NodeScan(a :: :A NODE, SolvedQueryModel(Set(),Set(a:A :: BOOLEAN),Set()))
· · · · · · · · |-SetSourceGraph(LogicalExternalGraph(c149c1b1-3887-4707-9ac8-b9d51f3519dd)(session:///graphs/ambient/c149c1b1-3887-4707-9ac8-b9d51f3519dd), SolvedQueryModel(Set(),Set(),Set()))
· · · · · · · · · |-Start(LogicalExternalGraph(c149c1b1-3887-4707-9ac8-b9d51f3519dd)(session:///graphs/ambient/c149c1b1-3887-4707-9ac8-b9d51f3519dd), Set(), SolvedQueryModel(Set(),Set(),Set()))
· · · · · |-NodeScan(c :: :C NODE, SolvedQueryModel(Set(),Set(c:C :: BOOLEAN),Set()))
· · · · · · |-Start(LogicalExternalGraph(c149c1b1-3887-4707-9ac8-b9d51f3519dd)(session:///graphs/ambient/c149c1b1-3887-4707-9ac8-b9d51f3519dd), Set(), SolvedQueryModel(Set(),Set(),Set()))
```
The node scan of `a` is simply overwritten by `b` instead of cross joined with it.

DarthMax:
fixes #270
fixes projection of non trivial list literals
fixes calculation of SparkSchema from RecordHeader
fixes evaluation of multiple OptionalMatch clauses
fixes conversion of CAPSNode/Rel from `null`


caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/ExpressionBehaviour.scala
```diff
@@ -615,6 +615,21 @@ trait ExpressionBehaviour {
...
       }
+
+      it("can convert ListLiterals with nested non literal expressions") {
+        val graph = initGraph("CREATE ({val: 1}), ({val: 2})")
+
+        val result = graph.cypher(
+          """
+            |MATCH (n)
+            |WITH [n.val*10, n.val*100] as vals
+            |RETURN val""".stripMargin)
```
Mats-SX: This doesn't look like valid Cypher. After the `WITH` horizon, only `vals` is available, so `RETURN val` should report an error; should be `RETURN vals`.

caps-spark/src/test/scala/org/opencypher/caps/impl/spark/acceptance/ExpressionBehaviour.scala
```diff
@@ -615,6 +615,21 @@ trait ExpressionBehaviour {
...
+
+        result.records.toMaps should equal(Bag(
+          CypherMap("vals" -> Seq(10, 100)),
+          CypherMap("val" -> Seq(20, 200))
```
Mats-SX: This doesn't look right; the columns can't change between rows. Should be `vals` on both rows after the Cypher query above is fixed.

caps-api/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -149,23 +129,9 @@ final case class Schema(
...
-        .map(nodeKeys)
-        .foldLeft(CTVoid: CypherType) {
-          case (inferred, next) => inferred.join(next.getOrElse(key, CTNull))
-        } match {
-        case CTNull | CTVoid => None
-        case tpe => Some(tpe)
-      }
+  def nodeKeyType(labels: Set[String], key: String): Option[CypherType] = {
+    val combos = combinationsFor(labels)
+    keysFor(combos).get(key)
```
Mats-SX: Nice!

caps-api/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
```diff
@@ -229,8 +195,6 @@ final case class Schema(
     * @return a copy of the Schema with the provided new data
     */
   def withNodePropertyKeys(nodeLabels: Set[String], keys: PropertyKeys): Schema = {
-    if (nodeLabels.exists(_.isEmpty))
-      throw SchemaException("Labels must be non-empty")
```
Mats-SX: Why did you remove this?

caps-ir/src/main/scala/org/opencypher/caps/ir/impl/typer/SchemaTyper.scala
```diff
@@ -72,9 +71,17 @@ object SchemaTyper {
         varTyp <- process[R](v)
         schema <- ask[R, Schema]
         result <- varTyp.material match {
+
+          // This means that the node can have any possible label combination, as the user did not specify any constraints
+          case n: CTNode if n.labels.isEmpty =>
```
Mats-SX: I think it should be possible to use `case CTNode(Set.empty)` here. If not, `case CTNode(labels) if labels.isEmpty` is slightly shorter.

caps-ir/src/test/scala/org/opencypher/caps/ir/impl/CypherQueryBuilderTest.scala
```diff
@@ -174,9 +174,11 @@ class CypherQueryBuilderTest extends IrTestSuite {
     }
   }
 
-  test("return graph of") {
+  it("can handle return graph of") {
     "MATCH (a), (b) RETURN GRAPH moo OF (a)-[r:TEST]->(b)".model.ensureThat { (model, globals) =>
-      val expectedSchema = Schema.empty.withRelationshipPropertyKeys("TEST")()
+      val expectedSchema = Schema.empty
+        .withNodePropertyKeys(Set.empty[String], PropertyKeys.empty)
+        .withRelationshipPropertyKeys("TEST")()
```
Mats-SX: Nice catch!

caps-api/src/main/scala/org/opencypher/caps/api/schema/Schema.scala
DarthMax: Prior to this commit we used a class `NoLabel` that simply inherited from Set[String] and was an empty set. We dropped that. Now NoLabel == Set.empty is what we use

Mats-SX: That seems sensible, but not related to this piece of code. This was a check that made sure that the empty string (an invalid label) is never accepted as input to the Schema. Without the check it would be possible to do `Schema.withNodePropKeys("", "key" -> CTInteger)`

DarthMax: Ohh, yes you are right. I read that line as `if(nodeLabels.isEmpty)`

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/convert/rowToCypherMap.scala
```diff
@@ -32,62 +33,72 @@ final case class rowToCypherMap(header: RecordHeader) extends (Row => CypherMap)
...
-        CypherValue(row.getAs[Any](SparkColumnName.of(s)))
-      }
-      .collect {
-        case (p, v) if !v.isNull =>
-          p.key.name -> v
-      }
+  private def collectNode(row: Row, field: Var): CypherValue = {
+    val idValue = row.getAs[Any](SparkColumnName.of(header.slotFor(field)))
+    idValue match {
+      case null       => CypherNull
```
Mats-SX: This is invalid; the id can never be `null`. We should probably throw an exception, but I think we already set these kinds of columns as non-nullable in Spark.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/convert/rowToCypherMap.scala
```diff
@@ -32,62 +33,72 @@ final case class rowToCypherMap(header: RecordHeader) extends (Row => CypherMap)
...
-        CypherValue.apply(row.getAs[Any](SparkColumnName.of(s)))
-      }
-      .collect {
-        case (p, v) if !v.isNull =>
-          p.key.name -> v
-      }
+  private def collectRel(row: Row, field: Var): CypherValue = {
+    val idValue = row.getAs[Any](SparkColumnName.of(header.slotFor(field)))
+    idValue match {
+      case null       => CypherNull
```
Mats-SX: Same comment as above; I think we should violently fail here.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/convert/rowToCypherMap.scala
DarthMax: It can, if we do an OptionalMatch

Mats-SX: Ah, you're right!

PR #280, opened 2018-02-07
Generalize and move physical planning to caps-cora

s1ck:


s1ck:
adding some documentation

PR #281, opened 2018-02-08
Add support for input driving table

Mats-SX:


Mats-SX:
Not sure what's up with GitHub; the commits are listed here in the wrong order compared to my local branch.

PR #282, opened 2018-02-08
Pom fixes

Mats-SX:
- Stops building shaded JAR by default
- Builds source JARs in `package` phase
- Formats all pom files with 2 space indent instead of 4 (saves horizontal space)
- Removes some strange executions from the pom

To reviewer: you can skip looking at the last commit (GitHub allows reviewing only a subset of the commits) as it contains only whitespace changes.


pstutz:
Nice, I like the simplifications.

Mats-SX:
Ah, that was an oversight. I'll try and restore it.

Mats-SX:
I pushed the `-Xfatal-warnings` directly: https://github.com/opencypher/cypher-for-apache-spark/commit/b07f15a15db34d0074f918d351da649d4290fb0b

PR #283, opened 2018-02-08
Prepare for CAPSRecords refactoring

pstutz:


caps-spark/src/main/scala/org/opencypher/caps/api/schema/EntityTable.scala
```diff
@@ -16,45 +16,118 @@
...
-  private[caps] def records(implicit caps: CAPSSession): CAPSRecords = CAPSRecords.create(this)
+}
+
+object EntityTable {
+
+  implicit class SparkTable(val df: DataFrame) extends CypherTable {
+
+    override def columns: Set[String] = df.columns.toSet
+
+    override def columnType: Map[String, CypherType] = columns.map(c => c -> SparkUtils.cypherTypeForColumn(df, c)).toMap
```
Mats-SX: Probably a bit premature, but this could probably be a `lazy val` to avoid redoing the column type conversion.

caps-spark/src/main/scala/org/opencypher/caps/api/schema/EntityTable.scala
```diff
@@ -76,25 +149,15 @@ case class NodeTable(mapping: NodeMapping, table: DataFrame) extends EntityTable
...
-    NodeTable(nodeMapping, nodeDF)
-  }
-
-  private def properties(nodeColumnNames: Seq[String]): Set[String] = {
-    nodeColumnNames.filter(_ != sourceIdKey).toSet
+    mapping.optionalLabelMapping.values.foreach { optionalLabelKey =>
+      val columnType = table.columnType(optionalLabelKey)
+      if (columnType != CTBoolean) {
+        throw IllegalArgumentException(
+          s"optional label key type in column `$optionalLabelKey`that is compatible with CTBoolean", columnType)
```
Mats-SX: I believe that this is unreachable. `table.columnType` will do a conversion from Spark's column type and fail internally if it can't. But perhaps this is an aide for some other implementation where we don't know the internal conversion scheme?

caps-spark/src/main/scala/org/opencypher/caps/api/schema/EntityTable.scala
```diff
@@ -113,44 +176,30 @@ case class RelationshipTable(mapping: RelationshipMapping, table: DataFrame) ext
...
-      relationshipType,
-      relationshipProperties)
+    val sourceStartNodeKeyType = table.columnType(mapping.sourceStartNodeKey)
+    if (sourceStartNodeKeyType != CTInteger) throw IllegalArgumentException(
+      s"start node key type in column `${mapping.sourceStartNodeKey}` that is compatible with CTInteger", sourceStartNodeKeyType)
 
-    RelationshipTable(relationshipMapping, relationshipDF)
-  }
+    val sourceEndNodeKeyType = table.columnType(mapping.sourceEndNodeKey)
+    if (sourceEndNodeKeyType != CTInteger) throw IllegalArgumentException(
```
Mats-SX: Actually, if this is supposed to be generic code, the type of the ID isn't fixed to integer. That is parameterised on the `Node` and `Relationship` types, right?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvGraphLoader.scala
```diff
@@ -125,11 +141,14 @@ class CsvGraphLoader(fileHandler: CsvGraphLoaderFileHandler)(implicit capsSessio
         optionalLabels = schema.optionalLabels.map(_.name).toSet,
         propertyKeys = schema.propertyFields.map(_.name).toSet)
 
-      NodeTable(nodeMapping, dataFrame)
+      val shouldBeNonNullable = schema.optionalLabels.map(_.name).toSet + schema.idField.name
```
s1ck: I know that the check needs to be done somewhere, but the EntityTable impls were the better place, imho. Maybe we somehow model that in again. Now each source (which can also be the user) needs to do the verification.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/convert/SparkUtils.scala
```diff
@@ -22,6 +22,20 @@ import org.opencypher.caps.api.types._
...
+  implicit class NullabilityOps(df: DataFrame) {
+    def setNonNullable(columnName: String): DataFrame = {
+      val newSchema = StructType(df.schema.map {
+        case s@StructField(cn, _, true, _) if cn == columnName => s.copy(nullable = false)
+        case other => other
+      })
+      if (newSchema == df.schema) {
+        df
+      } else {
+        df.sparkSession.createDataFrame(df.rdd, newSchema)
```
Mats-SX: This is a horizon operation, right? It cuts off the tree of DataFrame operators meaning no Catalyst optimisations across this boundary.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/hdfs/CsvGraphLoader.scala
pstutz: It's a bug-fix, not a check. So far these fields were all nullable, now an error is thrown when they are nullable, so they have to be set to non-nullable.

caps-spark/src/main/scala/org/opencypher/caps/api/schema/EntityTable.scala
pstutz: The check is more specific and asserts a particular type. So if the field is nullable or has the wrong Cypher type, then it fails here.

caps-spark/src/main/scala/org/opencypher/caps/api/schema/EntityTable.scala
pstutz: Yes, we should ensure that when we move the code to CORA. Another interesting constraint of the current model is that even though the ID can be something other than a Long, it has to be a CypherType (I think only CTString might make sense as an alternative).

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/convert/SparkUtils.scala
pstutz: Yes, it's only used to assign the correct nullability during CSV and DF import.

PR #289, opened 2018-02-12
Make timings output configurable

s1ck:


caps-cosc/src/main/scala/org/opencypher/caps/cosc/impl/planning/UnaryOperator.scala
```diff
@@ -57,24 +53,20 @@ case class Scan(in: COSCOperator, inGraph: LogicalGraph, v: Var, header: RecordH
...
-case class Filter(in: COSCOperator, expr: Expr, header: RecordHeader) extends COSCOperator {
+case class Filter(in: COSCOperator, expr: Expr, header: RecordHeader) extends UnaryOperator {
 
-  override def execute(implicit context: COSCRuntimeContext): COSCPhysicalResult = {
-    val prev = in.execute
-    val newRecords = prev.records
-    println(expr)
-    COSCPhysicalResult(newRecords, prev.graphs)
-  }
+  override def executeUnary(prev: COSCPhysicalResult)(implicit context: COSCRuntimeContext): COSCPhysicalResult = prev
```
DarthMax: The filter can not work if we do not print the expression :grin:

PR #290, opened 2018-02-12
Remove unnecessary indirection in SparkSQLExprMapper

s1ck:


caps-spark/src/main/scala/org/opencypher/caps/impl/spark/SparkSQLExprMapper.scala
```diff
@@ -98,18 +64,44 @@ object SparkSQLExprMapper {
...
+        // direct column lookup
+        case _: Var | _: Param | _: Property | _: HasLabel | _: StartNode | _: EndNode =>
+          verify
+
+          val slot = header.slotsFor(expr).head
+
+          val columns = df.columns.toSet
+          val colName = columnName(slot)
+
+          if (columns.contains(colName)) {
```
Mats-SX: This should never be false unless there's a bug, right?

Mats-SX: Or no, missing properties is a `false` case I suppose!

PR #292, opened 2018-02-13
Demonstrate access to result fields via Scala

s1ck:


README.md
```diff
@@ -128,12 +128,8 @@ object SocialNetworkData {
...
 The above program prints:
 ```
-+--------------------------------------------------------------------+
-| a.name               | b.name               | r.since              |
-+--------------------------------------------------------------------+
-| 'Alice'              | 'Bob'                | '23/01/1987'         |
-| 'Bob'                | 'Carol'              | '12/12/2009'         |
-+--------------------------------------------------------------------+
+Map(a.name -> Alice, b.name -> Bob, r.since -> 23/01/1987)
+Map(a.name -> Bob, b.name -> Carol, r.since -> 12/12/2009)
```
Mats-SX: Why is this nicer than the table printout?

s1ck: We had the same thought, it's not nicer. However, this example should show how to use the output data, e.g. by converting it to a map.

s1ck: Mh, but since `results.records.iterator` is easy to discover and the changes in `DataFrameExample` are showing the same behavior in an even more specific way, we should maybe revert this. It's also a nice analogy to what users know from Neo4j.

Mats-SX: Maybe we should do both? 4) Convert to maps like this: ... 5) Or just print a pretty table like this: ...

PR #294, opened 2018-02-13
Implement evaluation of distinct collect and count

DarthMax:
Allows evaluation of

```
MATCH (a)-->(b)
RETURN a,
                COLLECT(DISTINCT v.val) as list,
                COUNT(DISTINCT v.val2) as count
```

Distinct evaluation of average and sum are not supported by spark


caps-ir/src/test/scala/org/opencypher/caps/ir/impl/ExpressionConverterTest.scala
```diff
@@ -81,7 +81,7 @@ class ExpressionConverterTest extends BaseTestSuite with Neo4jAstTestSupport wit
 
   test("convert count") {
     convert(parseExpr("count(a)")) should equal(
-      Count(Var("a")())()
+      Count(Var("a")(), false)()
```
Mats-SX: Would be nice with the converse test for `count(DISTINCT a)` that results in a `true`.

PR #296, opened 2018-02-13
DF operation and table validation improvements

### LINKED ISSUE #295: Improve error message when the type for an id/label/relType column is nullable
Thanks to Ricky for reporting this. Currently the error message simply says that the type does not match, and the question mark at the end of the type is the only indication that the mismatch is due to nullability.

pstutz:
Fixes #295

Also several minor refactorings, such as a more consistent use of safe DF operations. This makes it easier to diagnose errors, because some mistakes result in errors as soon as an unintended operation is executed (instead of later during record header verification).


caps-spark/src/main/scala/org/opencypher/caps/impl/spark/DataFrameOps.scala
```diff
@@ -0,0 +1,208 @@
...
+      */
+    def structFieldForColumn(columnName: String): StructField = {
+      if (df.schema.fieldIndex(columnName) < 0) {
+        throw IllegalArgumentException(s"column with name $columnName", s"columns with names ${df.columns.mkString("[", ", ", "]")}")
+      }
+      df.schema.fields(df.schema.fieldIndex(columnName))
+    }
+
+    def mapColumn(columnName: String)(f: Column => Column): DataFrame = {
+      val tmpColName = ColumnName.tempColName
```
DarthMax: I think you can simply replace the column, e.g.
`.withColumn("existingColumn", data.col("existingColumn") + 2)`

this will replace the column value with itself + the expr

pstutz: Thanks, that's a nice simplification, I fixed it in the follow-up PR https://github.com/opencypher/cypher-for-apache-spark/pull/297

PR #297, opened 2018-02-13
API improvements and simplifications

pstutz:
Builds on top of https://github.com/opencypher/cypher-for-apache-spark/pull/296


caps-cosc/src/main/scala/org/opencypher/caps/cosc/impl/Demo.scala
```diff
@@ -21,7 +21,7 @@ import org.opencypher.caps.cosc.impl.value.{COSCNode, COSCRelationship}
 
 object Demo extends App {
 
-  PrintTimings.set()
+  PrintTimings.set
```
Mats-SX: I think Scala conventions say that one should include empty parentheses when the call has side effects.

pstutz: Thanks, will fix that.

pstutz:
Good point about the runtime allocations for the DataFrameOps implicit class. I'll extend AnyVal to get rid of them.

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/DataFrameOps.scala
```diff
@@ -91,7 +91,7 @@ object DataFrameOps {
     }
   }
 
-  implicit class RichDataFrame(df: DataFrame) {
+  implicit class RichDataFrame(val df: DataFrame) extends AnyVal {
```
Mats-SX: What does this actually do? If it removes the class instantiation, where does the field end up? On the containing object?

pstutz: There are only method calls and no fields on the wrapper. It exists only at compile time. Essentially the methods of the wrapper are inlined into the call-site by the compiler.
_"One use case for value classes is to combine them with implicit classes (SIP-13) for allocation-free extension methods."_ https://docs.scala-lang.org/overviews/core/value-classes.html

Mats-SX: Thanks for the pointer. I understand that it gets moved to a static object, which is cheap for the compiler -- that's a nice conversion because too many static objects typically litter the code at compile time.

PR #298, opened 2018-02-14
Add module containing examples for Cypher on Spark

s1ck:


s1ck:
The build fails for one profile due to exceeded build time. Could be reviewed and merged.

caps-examples/src/main/scala/org/opencypher/spark/examples/PageRankExample.scala
```diff
@@ -0,0 +1,83 @@
...
+  // 4) Create GraphX compatible RDDs from nodes and relationships
+  val graphXNodeRDD = nodes.records.asDF.rdd.map(row => (row.getLong(0), row.getString(1)))
+  val graphXRelRDD = rels.records.asDF.rdd.map(row => Edge(row.getLong(0), row.getLong(1), ()))
+
+  // 5) Compute Page Rank via GraphX
+  val graph = Graph(graphXNodeRDD, graphXRelRDD)
+  val ranks = graph.pageRank(0.0001).vertices.join(graphXNodeRDD).map { case (_, (rank, name)) => name -> rank }
+
+  // 6) Sort and print resulting ranks
+  val rankDF = session.sparkSession.createDataFrame(ranks)
```
Mats-SX: Just noting that re-wrapping this into a DataFrame could be potentially expensive.

caps-examples/src/main/scala/org/opencypher/spark/examples/PageRankExample.scala
```diff
@@ -0,0 +1,83 @@
...
+package org.opencypher.spark.examples
+
+import com.esotericsoftware.kryo.Kryo
+import org.apache.spark.graphx._
+import org.apache.spark.util.collection.{BitSet, OpenHashSet}
+import org.opencypher.caps.api.CAPSSession
+import org.opencypher.caps.api.CAPSSession._
+import org.opencypher.caps.impl.spark.CypherKryoRegistrator
+
+object PageRankExample extends App {
```
Mats-SX: I would like to have `GraphX` somewhere in the name of this example, perhaps `GraphXPageRankExample`?

It would also be nice with a description, perhaps:

```
/**
  * Demonstrates how CAPS results can be used to construct a GraphX graph and invokes a GraphX algorithm on it.
 */
```

caps-spark/src/main/scala/org/opencypher/caps/api/CAPSSession.scala
```diff
@@ -103,6 +103,42 @@ object CAPSSession extends Serializable {
...
+    * // ...
+    * val df: DataFrame = results.records.asDF
+    * }}}
+    *
+    */
+  implicit class RecordsAsDF(val records: CypherRecords) extends AnyVal {
+    /**
+      * Extracts the underlying [[DataFrame]] from the given [[records]].
+      */
+    def asDF: DataFrame = records match {
```
Mats-SX: I think we should add something on the doc here saying we have no guarantees on what the column names are (consider labels etc for flattened entities).

caps-spark/src/test/scala/org/opencypher/caps/demo/GCDemoTest.scala
```diff
@@ -105,7 +105,7 @@ class GCDemoTest extends CAPSTestSuite with SparkSessionFixture with Neo4jServer
     System.out.println(s"${tx - t0} ms")
   }
 
-  test("write back to Neo") {
+  ignore("write back to Neo") {
```
Mats-SX: Was this re-ignored due to the `netty` dependency move?

caps-examples/src/main/scala/org/opencypher/spark/examples/DataFrameOutputUsingAliasExample.scala
```diff
@@ -0,0 +1,65 @@
...
+  // 3) Query graph with Cypher
+  val results = socialNetwork.cypher(
+    """|MATCH (a:Person)-[r:FRIEND_OF]->(b)
+       |RETURN a.name, b.name, r.since""".stripMargin)
+
+  // 4) Extract DataFrame representing the query result
+  val df: DataFrame = results.records.asDF
+
+  // 5) Select specific return items from the query result
+  val projection: DataFrame = df.select(columnFor("a.name"), columnFor("b.name"))
```
Mats-SX: What happens if the query had returned a node `a` and we do `df.select(columnFor("a"))` here? Does it only give the id field back?

Perhaps we should offer two "drop-down" paths: 
- one that goes to `DataFrame` and detaches all Cypher notions from it; you just get flat entities and have to deal with that on your own (using RecordHeader or whatever)
- and one that goes to `Dataset[CypherMap]` with collected entities, which is type-safe and nice and stays within Catalyst, but is not Spark SQL

caps-examples/src/main/scala/org/opencypher/spark/examples/DataFrameOutputUsingAliasExample.scala
```diff
@@ -0,0 +1,65 @@
...
+  val results = socialNetwork.cypher(
+    """|MATCH (a:Person)-[r:FRIEND_OF]->(b)
+       |RETURN a.name AS person1, b.name AS person2, r.since AS friendsSince""".stripMargin)
+
+  // 4) Extract DataFrame representing the query result
+  val df: DataFrame = results.records.asDF
+
+  // 5) Select aliased return items from the query result
+  val projection: DataFrame = df
+    .select("person1", "friendsSince", "person2")
```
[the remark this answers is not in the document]
s1ck: Maybe it would be nice to guarantee that aliased return items are aligned with column names? If not, we should remove the whole example as it demonstrated that exact behavior. wdyt?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/CAPSSessionImpl.scala
```diff
@@ -48,6 +49,7 @@ sealed class CAPSSessionImpl(val sparkSession: SparkSession, private val graphSo
 
   self =>
 
+  private implicit def capsSession = this
```
Mats-SX: I did this change too in some commit -- I guess it didn't end up being merged.

caps-examples/src/main/scala/org/opencypher/spark/examples/PageRankExample.scala
```diff
@@ -42,19 +45,46 @@ object PageRankExample extends App {
...
+  // 9) Query across both graphs to print names with corresponding ranks, sorted by rank
+  val result = session.cypher(
+    """|FROM GRAPH AT '/ranks'
+       |MATCH (r)
+       |WITH id(r) as id, r.rank as rank
+       |FROM GRAPH AT '/sn'
+       |MATCH (p:Person)
+       |WHERE id(p) = id
+       |RETURN p.name as name, rank
+       |ORDER BY rank DESC""".stripMargin)
```
Mats-SX: Cool example!

caps-spark/src/test/scala/org/opencypher/caps/demo/GCDemoTest.scala
pstutz: It's not related to the dependency move.

We ignored it again because the test takes too long and if it were to test a feature, then that would better be done in a test specific to that feature and not in the demo test.

caps-examples/src/main/scala/org/opencypher/spark/examples/DataFrameOutputUsingAliasExample.scala
pstutz: We'll tackle this in a separate PR.

PR #301, opened 2018-02-16
Add collect method to CypherRecords

DarthMax:
During the benchmarks we found that `DataFrame.toLocalIterator` often is considerably slower than `collect`. This PR introduces collect on the public CypherRecord interfaces and uses the where possible

### DIFF
--- a/caps-api/src/main/scala/org/opencypher/caps/api/table/CypherRecords.scala
+++ b/caps-api/src/main/scala/org/opencypher/caps/api/table/CypherRecords.scala
@@ -27,7 +27,7 @@ trait CypherRecords extends CypherTable[String] with CypherPrintable {
   /**
     * Consume these records as an iterator.
     *
-    * WARNING: This operation may be very expensive as it may have to materialise
+    * WARNING: This operation may be very expensive as it may have to materialise the full result set.
     *
     * @note This method may be considerably slower than [[org.opencypher.caps.api.table.CypherRecords#collect]].
     *       Use this method only if collect could outgrow the available driver memory.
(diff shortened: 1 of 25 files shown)


caps-api/src/main/scala/org/opencypher/caps/api/table/CypherRecords.scala
```diff
@@ -28,9 +28,19 @@ trait CypherRecords extends CypherTable[String] with CypherPrintable {
...
+    *
+    * @note This method may be considerably slower than [[org.opencypher.caps.api.table.CypherRecords#collect]].
+    *       Use this method only if collect could outgrow the available driver memory.
     */
   def iterator: Iterator[CypherMap]
 
+  /**
+    * Consume these records and collect them into an array.
+    *
+    * WARNING: This operation may be very expensive as it may have to materialise
```
Mats-SX: ... the full result set?

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/CAPSRecords.scala
```diff
@@ -142,14 +142,14 @@ sealed abstract class CAPSRecords(val header: RecordHeader, val data: DataFrame)
   }
 
   def toLocalIterator: java.util.Iterator[CypherMap] = {
-    toCypherMaps.toLocalIterator()
+    toCypherMaps.collect().toIterator.asJava
```
Mats-SX: Is this faster than `.toLocalIterator`?

PR #303, opened 2018-02-20
Replace URI based data source handling with Catalog 

s1ck:


s1ck:
`CGDemoTest` still needs to be adapted

caps-logical/src/main/scala/org/opencypher/caps/logical/impl/LogicalPlanner.scala
```diff
@@ -346,16 +347,17 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
 
         LogicalPatternGraph(name, schema, GraphOfPattern(entities, boundEntities))
 
-      case _ =>
-        val graphSource = context.resolver(graph.name)
-        val schema = graphSource.schema match {
+      case g: IRQualifiedGraph =>
+        val graphSource = context.resolver(g.name)
+        // TODO can we use the schema attached to the IRGraph?
```
Mats-SX: I think so, yes. We added that schema through resolving the datasource during IR building. The reason we have a resolver also during logical planning is because we had the concept of scoped, named graphs where we only had a name at the reference point and no address -- this concept is going away. Now we will always be able to use the name to look up the graph (and its schema) from the catalog/cache.

caps-api/src/test/scala/org/opencypher/caps/api/graph/CypherSessionTest.scala
```diff
@@ -0,0 +1,90 @@
...
+    val session = createSession
+    val namespace = Namespace.from("foo")
+    val dataSource = mock[PropertyGraphDataSource]
+    session.registerSource(namespace, dataSource)
+    session.dataSource(namespace) should equal(dataSource)
+    session.deregisterSource(namespace)
+    an[IllegalArgumentException] should be thrownBy session.dataSource(namespace)
+  }
+
+  test("namespaces") {
```
Mats-SX: Nice tests!

caps-spark/src/main/scala/org/opencypher/caps/impl/spark/io/file/FileCsvPropertyGraphDataSource.scala
```diff
@@ -37,7 +37,8 @@ class FileCsvPropertyGraphDataSource(rootPath: String)(implicit val session: CAP
 
   override def store(name: GraphName, graph: PropertyGraph): Unit = ???
 
-  override def delete(name: GraphName): Unit = ???
+  override def delete(name: GraphName): Unit =
+    if (hasGraph(name)) Files.delete(Paths.get(graphPath(name)))
```
Mats-SX: Sweet :)

caps-examples/src/main/scala/org/opencypher/spark/examples/MultipleGraphExample.scala
```diff
@@ -30,9 +32,9 @@ object MultipleGraphExample extends App {
   val socialNetwork = session.readFrom(SocialNetworkData.persons, SocialNetworkData.friendships)
 
   // 3) Load purchase network data via local CSV + Schema files
-  val csvFolder = getClass.getResource("/csv/prod/").getFile
-  // Note: if files were stored in HDFS, change the URI scheme to `hdfs+csv` to load from HDFS instead
-  val purchaseNetwork = session.readFrom(s"file+csv://$csvFolder")
+  val csvFolder = getClass.getResource("/csv").getFile
+  // Note: if files were stored in HDFS, change the data source to HdfsCsvPropertyGraphDataSource
+  val purchaseNetwork = new FileCsvPropertyGraphDataSource(rootPath = csvFolder).graph(GraphName.from("prod"))
```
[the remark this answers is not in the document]
s1ck: Since the graph is used for a `union` in the next line, we need it right after. However, I changed it to `registerSource` followed by `seesion.graph(...)` to demonstrate the `QualifiedGraphName`

PR #306, opened 2018-02-24
Resolve several caps-api TODOs

s1ck:


caps-api/src/main/scala/org/opencypher/caps/api/graph/GraphOperations.scala
```diff
@@ -0,0 +1,34 @@
...
+
+object GraphOperations {
+
+  /**
+    * Converts a [[PropertyGraph]] into a [[PropertyGraph]] with [[GraphOperations]] if it is supported.
+    *
+    * @param graph property graph
+    * @return property graph with graph operations
+    */
+  implicit def withOperations(graph: PropertyGraph): PropertyGraph with GraphOperations = graph match {
```
s1ck: I wanted to leave the API unchanged and still allow writing `graph1 union graph2`. Alternatively, `CypherResult` could have a `type` field for the return type which is set by the implementing classes to something that might have `GraphOptions`. wdyt?

DarthMax: Mhh I like your second option more. Otherwise we could also use `asCAPS` and would thus not add another implicit conversion of PropertyGraphs.

Mats-SX: Wouldn't the second option require that the implementing class be known at compile time? Ie, that the API leaks the implementation?

caps-api/src/main/scala/org/opencypher/caps/api/configuration/Configuration.scala
```diff
@@ -15,33 +15,14 @@
...
-  abstract class ConfigFlag(name: String, defaultValue: Boolean = false)
-    extends ConfigOption[Boolean](name, defaultValue)(s => Try(s.toBoolean).toOption) {
-    def set(): Unit = set(true.toString)
-    def isSet = get
-  }
-
-  // TODO: Document
+  /**
+    * If enabled, the time required for executing a code block wrapped in
+    * [[org.opencypher.caps.impl.util.Measurement.time()]] will be printed.
```
Mats-SX: It doesn't say _which_ code block is timed, only _how_ it is timed; perhaps we should have both.

caps-api/src/main/scala/org/opencypher/caps/api/graph/CypherResult.scala
```diff
@@ -46,6 +46,8 @@ trait CypherResult extends CypherPrintable {
     * @return a table of records.
     */
   // TODO: Make option
+  // TODO: let's discuss this again (shouldn't the user know what the result is?)
+  // TODO: maybe provide an optRecords: Option[CypherRecords] in addition?
```
Mats-SX: Perhaps we should go Scala-esque and do `.asRecords` and `.asGraph` in addition to the options?

caps-api/src/main/scala/org/opencypher/caps/api/graph/GraphOperations.scala
```diff
@@ -0,0 +1,51 @@
...
+  self: PropertyGraph =>
+
+  /**
+    * Constructs the union of this graph and the argument graph. Note that, the argument graph has to be managed by the
+    * same session as this graph.
+    *
+    * @param other argument graph with which to union
+    * @return union of this and the argument graph
+    */
+  def union(other: PropertyGraph): PropertyGraph
```
Mats-SX: There's an additional `TODO` here for explaining the semantics of the union. There's another TODO to make sure we uphold those semantics.

caps-api/src/main/scala/org/opencypher/caps/api/graph/QualifiedGraphName.scala
```diff
@@ -0,0 +1,84 @@
...
+ * limitations under the License.
+ */
+package org.opencypher.caps.api.graph
+
+import org.opencypher.caps.api.io.PropertyGraphDataSource
+import org.opencypher.caps.impl.exception.IllegalArgumentException
+import org.opencypher.caps.impl.io.SessionPropertyGraphDataSource
+
+/**
+  * A graph name is used to address a specific graph withing a [[Namespace]] and is used for lookups in the
```
Mats-SX: Typo: within`g`

caps-api/src/main/scala/org/opencypher/caps/api/graph/GraphOperations.scala
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+  * Mixin for adding additional operations to a [[PropertyGraph]].
+  */
+trait GraphOperations {
+
+  self: PropertyGraph =>
+
+  /**
+    * Constructs the union of this graph and the argument graph. Note that, the argument graph has to be managed by the
```
Mats-SX: No comma after `Note that`

caps-api/src/main/scala/org/opencypher/caps/api/graph/QualifiedGraphName.scala
```diff
@@ -0,0 +1,84 @@
...
+    * Returns a [[QualifiedGraphName]] from its string representation. A qualified graph name consists of a namespace
+    * part and a graph name part separated by a '.' character. For example,
+    *
+    * {{{
+    *   mynamespace.mygraphname
+    *   mynamespace.my.graph.name
+    * }}}
+    *
+    * are valid qualified graph names. The separation between namespace and graph name is expected to be at the first
+    * occurring ','. A graph name may contain an arbitrary number of additional '.' characters. Note, that a string
```
Mats-SX: the first occurring '.' (full stop, not comma)

Mats-SX: No comma after `Note` either

caps-api/src/main/scala/org/opencypher/caps/api/io/PropertyGraphDataSource.scala
```diff
@@ -51,8 +51,8 @@ trait PropertyGraphDataSource {
     * Returns the [[Schema]] of the graph that is stored under the given name.
     *
     * This method gives implementers the ability to efficiently retrieve a graph schema from the data source directly.
-    * If an efficient retrieval is not possible, the call is typically forwarded to the graph using the
-    * [[PropertyGraph.schema]] call.
+    * For reasons of performance, it is highly recommended to make a schema available through this call. If an efficient
+    * retrieval is not possible, the call is typically forwarded to the graph using the [[PropertyGraph.schema]] call.
```
Mats-SX: I would expand with
> the [[PropertyGraph.schema]] call, which may require materialising the full graph.

caps-api/src/main/scala/org/opencypher/caps/impl/io/SessionPropertyGraphDataSource.scala
```diff
@@ -30,7 +30,7 @@ class SessionPropertyGraphDataSource() extends PropertyGraphDataSource {
 
   override def graph(name: GraphName): PropertyGraph = graphMap(name)
 
-  override def schema(name: GraphName): Option[Schema] = Some(graphMap(name).schema)
+  override def schema(name: GraphName): Option[Schema] = if (hasGraph(name)) Some(graph(name).schema) else None
```
Mats-SX: This seems a bit deceptive to me. Here we're using the `Option` as a means to protect ourselves from the graph (name) not being present and returning `None` when it isn't. The design indicated that `None` was to be used when no efficient retrieval was possible, which is different. This may be potentially confusing, as users may build logic that acts on this returning `None` as the graph was not present, which wouldn't match how other data source implementations meant that value to be interpreted.

caps-api/src/test/scala/org/opencypher/caps/impl/io/SessionPropertyGraphDataSourceTest.scala
```diff
@@ -15,29 +15,63 @@
...
 class SessionPropertyGraphDataSourceTest extends FunSuite with MockitoSugar with Matchers {
 
   test("hasGraph should return true for existing graph") {
     val source = new SessionPropertyGraphDataSource
-    val testGraphName = GraphName.from("test")
+    val testGraphName = GraphName("test")
     source.store(testGraphName, null)
-    source.hasGraph(testGraphName) should be(true)
+    source.hasGraph(testGraphName) should be
+    true
```
Mats-SX: Why is this on its own line?

caps-api/src/test/scala/org/opencypher/caps/impl/io/SessionPropertyGraphDataSourceTest.scala
```diff
@@ -15,29 +15,63 @@
...
   }
 
   test("hasGraph should return false for non-existing graph") {
     val source = new SessionPropertyGraphDataSource
-    val testGraphName = GraphName.from("test")
-    source.hasGraph(testGraphName) should be(false)
+    val testGraphName = GraphName("test")
+    source.hasGraph(testGraphName) should be
+    false
```
Mats-SX: Ditto own line

caps-api/src/test/scala/org/opencypher/caps/impl/io/SessionPropertyGraphDataSourceTest.scala
```diff
@@ -15,29 +15,63 @@
...
+    val source = new SessionPropertyGraphDataSource
+    val testGraphName = GraphName("test")
+    an[NoSuchElementException] should be thrownBy source.graph(testGraphName)
+  }
+
+  test("schema should return None for non-existing graph") {
+    val source = new SessionPropertyGraphDataSource
+    val testGraphName = GraphName("test")
+    source.schema(testGraphName).isEmpty should be
+    true
```
Mats-SX: And here

PR #307, opened 2018-02-26
Module and package renamings

Mats-SX:


okapi-logical/pom.xml
```diff
@@ -22,29 +22,29 @@
     <!--CAPS Module-->
     <dependency>
       <groupId>org.opencypher</groupId>
-      <artifactId>caps-api</artifactId>
+      <artifactId>okapi-api</artifactId>
```
Mats-SX: Don't think we need this dependency as we're depending on `ir` below

PR #316, opened 2018-03-02
Schema trait

DarthMax:
- Separate Schema and schema implementation
- new schema implementation for CAPS that verifies SPARK compatibility

### DIFF
--- a/okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
+++ b/okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
@@ -1,0 +1,269 @@
+/*
+ * Copyright (c) 2016-2018 "Neo4j, Inc." [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.opencypher.okapi.impl.schema
+
+import org.opencypher.okapi.api.schema.PropertyKeys.PropertyKeys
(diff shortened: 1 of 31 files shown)


okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
```diff
@@ -0,0 +1,269 @@
...
+ */
+package org.opencypher.okapi.impl.schema
+
+import org.opencypher.okapi.api.schema.PropertyKeys.PropertyKeys
+import org.opencypher.okapi.api.schema.{LabelPropertyMap, PropertyKeys, RelTypePropertyMap, Schema}
+import org.opencypher.okapi.api.types._
+import org.opencypher.okapi.impl.exception.SchemaException
+import org.opencypher.okapi.impl.schema.SchemaUtils._
+
+final case class SchemaImpl(labelPropertyMap: LabelPropertyMap, relTypePropertyMap: RelTypePropertyMap) extends Schema {
```
Mats-SX: Nice removing the other inputs

okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
```diff
@@ -0,0 +1,269 @@
...
+    // Map over the rest of the existing keys to mark them all nullable
+    val propertiesMarkedOptional = existing.filterKeys(k => !input.contains(k)).foldLeft(keysWithJoinedTypes) {
+      case (map, (key, propTyp)) =>
+        map.updated(key, propTyp.nullable)
+    }
+
+    propertiesMarkedOptional
+  }
+
+  override def withRelationshipPropertyKeys(typ: String, keys: PropertyKeys): Schema = {
```
Mats-SX: I guess we should check that `typ` isn't an empty string here

spark-cypher/src/main/scala/org/opencypher/spark/schema/CAPSSchema.scala
```diff
@@ -0,0 +1,119 @@
...
+import org.opencypher.spark.impl.convert.CAPSCypherType._
+
+object CAPSSchema {
+  val empty: CAPSSchema = Schema.empty.asCaps
+
+  implicit class CAPSSchemaConverter(schema: Schema) {
+
+    /**
+      * Converts a given schema into a CAPS specific schema. The conversion fails if the schema specifies property types
+      * that cannot be represented in CAPS and throws a [[org.opencypher.okapi.impl.exception.SchemaException]].
```
Mats-SX: It might be a bit strange to throw an okapi exception when this is a spark-specific implementation... but I don't mind

PR #318, opened 2018-03-07
Implement store for CSV based data sources

s1ck:


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -212,12 +212,6 @@ final case class Property(m: Expr, key: PropertyKey)(val cypherType: CypherType
   override def hashCode(): Int = {
     MurmurHash3.orderedHash(Seq(m, key, cypherType), MurmurHash3.stringHash(productPrefix))
   }
-
-  override def equals(obj: scala.Any) = obj match {
```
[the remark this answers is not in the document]
pstutz: Maybe also use the generated case class hashCode implementation?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
```diff
@@ -169,47 +168,52 @@ sealed abstract class CAPSRecords(val header: RecordHeader, val data: DataFrame)
...
-    val alignedData = this
-      .toDF()
-      .map { (row: Row) =>
-        val alignedRow = slotDataSelectors.map(_ (row))
-        new GenericRowWithSchema(alignedRow.toArray, wrappedHeader.asSparkSchema).asInstanceOf[Row]
-      }(wrappedHeader.rowEncoder)
 
-    CAPSRecords.verifyAndCreate(targetHeader, alignedData)
+    val withRelevantColumns = withRenamedColumns.select(relevantColumns: _*)
+    CAPSRecords.verifyAndCreate(targetHeader, withRelevantColumns)
```
pstutz: Still quite a read but looks a lot more solid than before. I assume that it's intended that it crashes during the cast if that cast is not possible? That error might be difficult to reconcile with the type mismatch that caused it.

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -114,9 +114,13 @@ object SparkSQLExprMapper {
           exprs.map(_.asSparkSQLExpr).foldLeft(functions.lit(false))(_ || _)
 
         case In(lhs, rhs) =>
-          val element = lhs.asSparkSQLExpr
-          val array = rhs.asSparkSQLExpr
-          array_contains(array, element)
+          if (rhs.cypherType == CTNull || lhs.cypherType == CTNull) {
```
pstutz: There's no way to check if an array contains a null element?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSScanGraph.scala
```diff
@@ -96,6 +96,8 @@ class CAPSScanGraph(val scans: Seq[CAPSEntityTable], val schema: CAPSSchema)(imp
         .groupBy(_.entityType)
         .flatMap { case (k, entityScans) => NonEmptyVector.fromVector(entityScans).map(k -> _) }
 
+    def byExactType(entityType: EntityType): Seq[CAPSEntityTable] = entityTablesByType(entityType).toVector
```
pstutz: Can this return multiple tables (suggested by Seq return type)?

spark-cypher/src/main/scala/org/opencypher/spark/impl/io/hdfs/CsvFileHandler.scala
```diff
@@ -0,0 +1,106 @@
...
+  override def writeSchemaFile(directory: String, filename: String, jsonSchema: String): Unit = {
+    val hdfsPath = new Path(new Path(graphLocation.getPath, directory), filename)
+    val outputStream = fs.create(hdfsPath)
+    val bw = new BufferedWriter(new OutputStreamWriter(outputStream, "UTF-8"))
+    bw.write(jsonSchema)
+    bw.close()
+  }
+}
+
+final class LocalFileHandler(override val graphLocation: URI) extends CsvFileHandler {
```
pstutz: Might it be possible to share more code with the `HadoopFileHandler` by also using the Hadoop`FileSystem`? I read that that one also supports local file systems.

spark-cypher/src/test/scala/org/opencypher/spark/test/fixture/TeamDataFixture.scala
```diff
@@ -101,6 +97,39 @@ trait TeamDataFixture extends TestDataFixture with RowDebugOutputSupport {
     Row(30L, 3L, "KNOWS", 4L, 2015L)
   )
 
+  // TODO: remove once https://issues.apache.org/jira/browse/SPARK-23610 is resolved
```
pstutz: Nice bug report.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
DarthMax: The cast should always be possible since the schema already verified that the scans are compatible.

okapi-relational/src/test/scala/org/opencypher/okapi/relational/impl/table/RecordHeaderTest.scala
```diff
@@ -163,7 +163,8 @@ class RecordHeaderTest extends BaseTestSuite {
       ))
   }
 
-  test("concatenating headers with similar properties") {
+  // TODO: do we support this behaviour? If yes, Property#equals needs to be overwritten again
```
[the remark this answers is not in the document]
DarthMax: Since the equality also includes the owner of the property and a graph element cannot have the same property twice but with different types it does not really make sense to include the cypher type into the equality. Especially since we do not include it into the other expressions as well.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
DarthMax: I believe that there are only two possible paths 
1. cast from NullType into something else nullable -> always possible
2. cast into a less specific class (e.g String -> Any) -> always possible

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
DarthMax: Nope Spark fails with an error if one of the arguments to `array_contains` is `null` :/

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSScanGraph.scala
DarthMax: Yes, since there could be multiple scans expressing the same type.

spark-cypher/src/main/scala/org/opencypher/spark/impl/io/hdfs/CsvFileHandler.scala
DarthMax: That is how it was in the beginning. But we ran into some kinds of trouble (I forgot what it was) and so we decided to split it up.

okapi-relational/src/test/scala/org/opencypher/okapi/relational/impl/table/RecordHeaderTest.scala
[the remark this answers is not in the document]
pstutz: Makes sense, let's remove the test then.

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
pstutz: That's weird. I wonder if it makes sense for us to propagate that oddity or to do it with a UDF when the element is null. Should we leave it as is and add a `UDF for null element check` issue?

DarthMax: To support all variants of Cypher, we will have to introduce an UDF
Because in the case of `null in ()` we currently return null, but the TCK says it should be false.

Mats-SX: That's interesting, I wonder if the TCK is wrong on that particularity. Ie, whether `IN` should use tri-valued logic.

PR #330, opened 2018-03-22
Add MERGE / CLONE support in CONSTRUCT

s1ck:


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/IRElement.scala
```diff
@@ -59,4 +59,8 @@ object IRCatalogGraph {
 
 final case class IRCatalogGraph(qualifiedName: QualifiedGraphName, schema: Schema) extends IRGraph
 
-final case class IRPatternGraph[E](schema: Schema, creates: Pattern[E], sets: List[SetPropertyItem[E]]) extends IRGraph
+final case class IRPatternGraph[E](
+  schema: Schema,
+  clones: Pattern[E],
```
Mats-SX: I understand this is a pattern now, but it'll be a list of `ReturnItem`s soon. I guess we'll want to represent this as a `Map[Expr, Var]` or similar.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -164,15 +164,39 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
...
+            val clonePatternSchema = context.workingGraph.schema.forPattern(clonePattern)
+
+            // computing single nodes/rels constructed by NEW (CREATE)
+            val newPattern = newPatterns.foldLeft(Pattern.empty[Expr])(_ ++ _)
+            val fieldNamesInNewPattern = newPattern.fields.map(_.name)
+            val newPatternSchema = context.workingGraph.schema.forPattern(newPattern)
+
+            // compute SET items for NEW (CREATE) patterns
+            val patternSchema = newPatternSchema ++ clonePatternSchema
+            val fieldNamesInPattern = fieldNamesInClonePattern ++ fieldNamesInNewPattern
```
Mats-SX: Nice structure.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/test/support/DebugOutputSupport.scala
```diff
@@ -26,17 +26,41 @@
...
+        } else if (x == null) {
+          -1
+        } else if (y == null) {
+          1
+        } else {
+          Ordering[String].compare(x.toString + x.hashCode, y.toString + y.hashCode)
+        }
+      }
+    }
+    TreeMap(elements.groupBy(identity).mapValues(_.size).toSeq: _*)
```
Mats-SX: Is this implementing the Bag using an ordered collection? Why? Don't we want our tests to be agnostic to row order, if we were the change something?

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/MultipleGraphExample.scala
```diff
@@ -57,14 +59,16 @@ object MultipleGraphExample extends App {
...
+       |  MERGE (c)
        |  CREATE (p)-[x:IS]->(c)
        |}
        |RETURN GRAPH
     """.stripMargin
   ).getGraph
 
   // 6) Build recommendation graph from disconnected and integration graphs
-  val recommendationGraph = disconnectedGraph unionAll integrationGraph
+  val recommendationGraph = disconnectedGraph.asCaps union integrationGraph
```
Mats-SX: Is this equivalence UNION?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSPatternGraph.scala
```diff
@@ -92,7 +92,8 @@ class CAPSPatternGraph(
     val extractedDf = baseTable
       .toDF()
       .flatMap(RowExpansion(targetHeader, targetVar, extractionSlots, relColumnsLookupTables))(targetHeader.rowEncoder)
-    val distinctData = extractedDf.distinct()
+
+    val distinctData = extractedDf.dropDuplicates(ColumnName.of(OpaqueField(targetVar)))
```
Mats-SX: Is this the implementation for deduplicating `CLONE`d entities?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/IRElement.scala
pstutz: We're working on this on the other PR https://github.com/opencypher/cypher-for-apache-spark/pull/331

okapi-ir/src/test/scala/org/opencypher/okapi/ir/test/support/DebugOutputSupport.scala
pstutz: As soon as both are bags they're both ordered the same way and diffs show up nicely aligned below each other in the error output.

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/MultipleGraphExample.scala
pstutz: No, it's more like a union that does not do graph retagging and might have (intended) collisions due to that. It was mainly a hidden workaround to be able to test things whilst `CONSTRUCT ON` was not around yet.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSPatternGraph.scala
pstutz: Yes. Unfortunately the `UnionGraph` does it again on the scans ... and `DISTINCT` is expensive. So maybe we pay a high price for it, but I find it reasonably easy to understand and it  is easy to implement like this.

pstutz: It's certainly less expensive than checking for equivalence. :)

okapi-ir/src/test/scala/org/opencypher/okapi/ir/test/support/DebugOutputSupport.scala
Mats-SX: Aha, so the _written order_ in the test is not the order it will actually compare with, then. It will sort the test expectation and the actual result and then compare. As long as we keep this in mind it'll be fine, got it.

pstutz: Yes, it's merely a canonical representation for the comparison. It even helps when looking at diffs, because you'd like for those to be in the same order. Right now that representation is a Sorted map, with the element count in the value. Probably that wouldn't even be necessary and we could use a sorted list instead.

PR #332, opened 2018-03-23
Implement `CONSTRUCT ON`

pstutz:
Co-Authored-By: Martin Junghanns <s1ck@users.noreply.github.com>
Co-Authored-By: Mats Rydberg <mats@neotechnology.com>
Co-Authored-By: Max Kießling <max@kopfueber.org>

### DIFF
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -202,7 +202,7 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
               setItem match {
                 case SetLabelItem(variable, labels) =>
                   val existingLabels = rewrittenVarTypes.getOrElse(variable, variable.cypherType) match {
-                    case CTNode(existing) => existing
+                    case CTNode(existing, _) => existing
                     case other => throw UnsupportedOperationException(s"SET label on something that is not a node: $other")
                   }
                   val labelsAfterSet = existingLabels ++ labels
@@ -215,8 +215,10 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
               }
             }
             val onGraphs = on.map(graph => QualifiedGraphName(graph.parts))
+            val qgn = context.uniqueSessionGraphNameGenerator()
 
             val patternGraph = IRPatternGraph[Expr](
+              qgn,
               schema,
               cloneItemMap,
               newPattern,
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -33,6 +33,7 @@ import org.neo4j.cypher.internal.frontend.v3_4.ast
 import org.neo4j.cypher.internal.util.v3_4.InputPosition
 import org.neo4j.cypher.internal.v3_4.{expressions => exp}
 import org.opencypher.okapi.api.graph.QualifiedGraphName
+import org.opencypher.okapi.api.schema.Schema
 import org.opencypher.okapi.api.types._
 import org.opencypher.okapi.impl.exception.{IllegalArgumentException, IllegalStateException, UnsupportedOperationException}
 import org.opencypher.okapi.ir.api._
@@ -42,6 +43,7 @@ import org.opencypher.okapi.ir.api.pattern.Pattern
 import org.opencypher.okapi.ir.api.set.{SetItem, SetLabelItem, SetPropertyItem}
 import org.opencypher.okapi.ir.api.util.CompilationStage
 import org.opencypher.okapi.ir.impl.refactor.instances._
+import org.opencypher.okapi.impl.schema.TagSupport._
 
 object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBuilderContext] {
 
@@ -184,18 +186,20 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
             val cloneItemMap: Map[IRField, Expr] = cloneItems.toMap
             val fieldsInCloneItems = cloneItemMap.keys.toSet
             val fieldNamesInCloneItems = fieldsInCloneItems.map(_.name)
-            val clonePatternSchema = context.workingGraph.schema.forFields(fieldsInCloneItems)
 
             // computing single nodes/rels constructed by NEW (CREATE)
             val newPattern = newPatterns.foldLeft(Pattern.empty[Expr])(_ ++ _)
             val fieldNamesInNewPattern = newPattern.fields.map(_.name)
-            val newPatternSchema = context.workingGraph.schema.forPattern(newPattern)
+
+            // Fields inside of CONSTRUCT could have been matched on other graphs than just the workingGraph
+            val cloneSchema = schemaForFields(context, cloneItemMap.keys.toSet)
+            val newPatternSchema = schemaForPattern(context, newPattern)
 
             // compute SET items for NEW (CREATE) patterns
-            val patternSchema = newPatternSchema ++ clonePatternSchema
+            val patternSchema = newPatternSchema ++ cloneSchema
             val fieldNamesInPattern = fieldNamesInCloneItems ++ fieldNamesInNewPattern
 
-            val (schema, _) = setItems.foldLeft(patternSchema -> Map.empty[Var, CypherType]) { case ((currentSchema, rewrittenVarTypes), setItem: SetItem[Expr]) =>
+            val (constructedSchema, _) = setItems.foldLeft(patternSchema -> Map.empty[Var, CypherType]) { case ((currentSchema, rewrittenVarTypes), setItem: SetItem[Expr]) =>
               if (!fieldNamesInPattern.contains(setItem.variable.name)) {
                 throw UnsupportedOperationException("SET on a variable that is not defined inside of the CONSTRUCT scope")
               }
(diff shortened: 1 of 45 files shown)


pstutz:
Graph construction is now planned:
```
USE GRAPH socialNetwork
MATCH (p:Person)
USE GRAPH csv.prod
MATCH (c:Customer)
WHERE p.name = c.name
CONSTRUCT ON socialNetwork, csv.prod {
  CLONE p, c
  CREATE (p)-[:IS]->(c)
}
RETURN GRAPH
```
=>

Physical plan:
```
|-ReturnGraph
· |-GraphUnionAll(false)
· · |-GraphUnionAll(true)
· · · |-Start(CAPSScanGraph(:Person NODE, :FRIEND_OF RELATIONSHIP))
· · · |-Start(CAPSScanGraph(:Customer NODE, :Product NODE, :BOUGHT RELATIONSHIP))
· · |-ConstructGraph(on=[session.socialNetwork, csv.prod], entities=[p :: :Person NODE @ socialNetwork, c :: :Customer NODE @ prod,   UNNAMED166 :: :IS RELATIONSHIP @ prod])
· · · |-Filter(p.name :: NULL = c.name :: STRING?, RecordHeader with 7 slots)
· · · · |-CartesianProduct(RecordHeader with 7 slots)
· · · · · |-UseGraph(LogicalCatalogGraph(qualifiedGraphName = csv.prod))
· · · · · · |-Scan(p :: :Person NODE, RecordHeader with 4 slots)
· · · · · · · |-Start(CAPSScanGraph(:Person NODE, :FRIEND_OF RELATIONSHIP))
· · · · · |-Scan(c :: :Customer NODE, RecordHeader with 3 slots)
· · · · · · |-Start(CAPSScanGraph(:Customer NODE, :Product NODE, :BOUGHT RELATIONSHIP))
```
That should make it easier to understand what's going on underneath. It also shows that the expression typer currently has trouble typing variables from a previous working graph (`p.name :: NULL`), so we'll need to use the graph type annotations for more than just pattern matching.

Another issue is `UNNAMED166 :: :IS RELATIONSHIP @ prod`. It seems that assigning patterns to the working graph by default is not correct and we should instead assign the QGN of the graph that is constructed by the pattern.

pstutz:
The current approach also has a weakness: In order to properly re-tag relationships we'll need to track what graph the start/end nodes are from. That's cumbersome, so I wonder if there's an alternative.

Mats-SX:
@pstutz Relationships aren't allowed to span graphs, so the tags used for their start/end nodes will be the same as for the graph of the relationship itself. I think we can use a uniform scheme for tags on nodes and relationships and not track these separately to keep things simple.

pstutz:
@Mats-SX When we create relationships in a `CONSTRUCT ON`, then it can refer to nodes that have different start/end tags. They're all in the same graph, but a graph can contain multiple tags.

pstutz:
Remaining issues:
- Move alias code to `CAPSRecords`
- Use physical planner to plan aliasing
- Track start/end QGN on relationship type (if we don't come up with an alternative approach)
- Improve typer so it can type variables from outside the scope of the working graph
- Track QGN/schema mappings of constructed graphs and access graph schemas in a way such that the schemas of constructed graphs an be resolved in the same way as catalog graphs
- Assign correct constructed graph QGN to new nodes created in `CONSTRUCT`

Mats-SX:
I'm not quite following. The start/end slots in the `RecordHeader` are typed as `CTNode` as well, so they would already contain a QGN reference and be treated the same as "normal" nodes would. Sure, they could have different tags between them, but that's not a more complicated if case than if they were separated vertically (across rows) instead of horizontally (across columns), is it?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -184,40 +186,55 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
...
                   val propertyType = setValue.cypherType
                   val updatedSchema = currentSchema.addPropertyToEntity(propertyKey, propertyType, variable.cypherType)
                   updatedSchema -> rewrittenVarTypes
               }
             }
+
             val onGraphs = on.map(graph => QualifiedGraphName(graph.parts))
+            val patternGraphSchema = {
+              if (onGraphs.size > 0) {
+                val onGraphsSchema = onGraphs.map(context.schemaFor).map(_.toTagged)
```
Mats-SX: Why are tags not on all schemas?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -184,40 +186,55 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherQuery[Expr], IRBu
...
+                val schemaForUnionOfOnGraphs = onGraphsSchema.reduce(_ union _)
+                schemaForUnionOfOnGraphs ++ constructedSchema // No UNION to avoid retagging
+              } else {
+                constructedSchema
+              }
+            }
+
+            val qgn = context.uniqueSessionGraphNameGenerator()
```
Mats-SX: This looks like a call to get the name generator, not a call to generate a new name. Could we rename this?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilderContext.scala
```diff
@@ -43,6 +43,7 @@ import org.opencypher.okapi.ir.impl.typer.exception.TypingException
 import org.opencypher.okapi.ir.impl.typer.{SchemaTyper, TypeTracker}
 
 final case class IRBuilderContext(
+  uniqueSessionGraphNameGenerator: () => QualifiedGraphName,
```
Mats-SX: Not a very nice signature, we should consider extracting this to a named trait or class

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/PatternConverter.scala
```diff
@@ -48,57 +49,61 @@ final class PatternConverter {
...
+          case None => FreshVariableNamer(np.position.offset, CTNode(allLabels, qgnOption))
         }
         for {
           entity <- pure(IRField(nodeVar.name)(nodeVar.cypherType))
           _ <- modify[Pattern[Expr]](_.withEntity(entity))
         } yield entity
 
-      case rc @ ast.RelationshipChain(left, ast.RelationshipPattern(eOpt, types, None, None, dir, _), right) =>
+      case rc
+        @ast.RelationshipChain(left, ast.RelationshipPattern(eOpt, types, None, None, dir, _), right)
```
Mats-SX: Weird line break

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/package.scala
```diff
@@ -89,14 +78,14 @@ package object impl {
 
     def addPropertyToEntity(propertyKey: String, propertyType: CypherType, entityType: CypherType): Schema = {
       entityType match {
-        case CTNode(labels) =>
+        case CTNode(labels, _) =>
```
Mats-SX: Since we most of the time seem to ignore this field, perhaps it signals that using a second parameter list would be a good idea.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -251,7 +251,7 @@ class IrBuilderTest extends IrTestSuite {
         |RETURN GRAPH""".stripMargin
 
     query.model.result match {
-      case GraphResultBlock(_, IRPatternGraph(schema,_, _, _, _)) =>
+      case GraphResultBlock(_, IRPatternGraph(_, schema, _, _, _, _)) =>
```
Mats-SX: It would be nice with some tests that checked that the graph name was set correctly for the clone entities, and also for entities in `NEW` patterns.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
```diff
@@ -62,19 +62,37 @@ class PhysicalPlanner[P <: PhysicalOperator[R, G, C], R <: CypherRecords, G <: P
...
+              case severalGraphs =>
+                val onGraphUnion = producer.planGraphUnionAll(severalGraphs.map(catalogGraph))
+                Some(onGraphUnion)
+            }
+
+            // If necessary plan a union if there are graphs that the new graph is CONSTRUCTED ON.
+            constructOnGraph match {
+              case None => constructGraphOperator
+              // Do not prevent collisions with the constructed graph in order to preserve the constructed relationships
+              case Some(constructedOn) => producer.planGraphUnionAll(constructedOn :: constructGraphOperator :: Nil, preventIdCollisions = false)
```
Mats-SX: The `:: Nil` syntax seems a bit off... how about `List(constructedOn, constructGraphOperator)`?

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/UnaryOperators.scala
```diff
@@ -198,37 +201,63 @@ final case class Filter(in: CAPSPhysicalOperator, expr: Expr, header: RecordHead
...
+    val retaggedBaseTableData = slotsToRetag.foldLeft(baseTable.data) { case (updatedTable, (slot, qgn)) =>
+      updatedTable.safeReplaceTags(ColumnName.of(slot), retaggings(qgn))
+    }
+    val retaggedBaseTable = CAPSRecords.verifyAndCreate(baseTable.header, retaggedBaseTableData)
+
     // Construct NEW entities
-    val newEntityTag = if (initialSchema.tags.isEmpty) 0 else initialSchema.tags.max + 1
-    val entityTable = createEntities(newItems, baseTable, newEntityTag)
-    val tableWithConstructedProperties = setProperties(setItems, entityTable)
+    val (newEntityTags, tableWithConstructedEntities) = {
```
DarthMax: new EntityTag does not need to be a Set

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/UnaryOperators.scala
```diff
@@ -198,37 +201,63 @@ final case class Filter(in: CAPSPhysicalOperator, expr: Expr, header: RecordHead
...
+      if (newEntities.isEmpty) {
+        Set.empty[Int] -> retaggedBaseTable
+      } else {
+        val newEntityTag = if (initialSchema.tags.isEmpty) 0 else initialSchema.tags.max + 1
```
DarthMax: maybe add a convenience method to TagSupport?

PR #335, opened 2018-03-28
Physical operator refactoring

s1ck:
This PR replaces some of the "graph specific" operators (e.g. `ExpandSource` and `ExpandInto`) as we;; as "special" operators (e.g. value join) and replaces them with pure relational operations (i.e. `Join` and `Selection/Filter`). This reduces complexity of the physical operators and requires less implementation effort for alternative relational backends. It also brings us closer to a pure relational output in `okapi-relational`. 

An example physical plan now looks like that:
```
|-SelectFields(List(nodes :: INTEGER), RecordHeader with 1 slots)
· |-Aggregate(Set((nodes :: INTEGER,count(b :: :B NODE @ tmp#15))), Set(), RecordHeader with 1 slots)
· · |-Join(List((target(  UNNAMED10 :: RELATIONSHIP @ tmp#15),b :: NODE @ tmp#15)), RecordHeader with 9 slots)
· · · |-Join(List((n :: NODE @ tmp#15,source(  UNNAMED10 :: RELATIONSHIP @ tmp#15))), RecordHeader with 7 slots)
· · · · |-NodeScan(n :: NODE, RecordHeader with 3 slots)
· · · · · |-Start(CAPSRecords.unit, org.opencypher.spark.impl.CAPSPatternGraph@388db4a6)
· · · · |-RelationshipScan(  UNNAMED10 :: RELATIONSHIP @ tmp#15, RecordHeader with 4 slots)
· · · · · |-Start(org.opencypher.spark.impl.CAPSPatternGraph@388db4a6)
· · · |-NodeScan(b :: :B NODE, RecordHeader with 2 slots)
· · · · |-Start(CAPSRecords.unit, org.opencypher.spark.impl.CAPSPatternGraph@388db4a6)
```

### DIFF
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
@@ -258,6 +258,17 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
   def planCartesianProduct(lhs: P, rhs: P, header: RecordHeader): P
 
   /**
+    * Joins the two input records on given variables.
+    *
+    * @param lhs         first previous operator
+    * @param rhs         second previous operator
+    * @param joinColumns sequence of left and right join columns
+    * @param header      resulting record header
+    * @return join operator
+    */
+  def planJoin(lhs: P, rhs: P, joinColumns: Seq[(Expr, Expr)], header: RecordHeader): P
+
+  /**
     * Joins the two input records on node attribute values.
     *
     * @param lhs        first previous operator
@@ -374,7 +385,7 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
   /**
     * Performs a UNION ALL over graphs.
     *
-    * @param graphs graph to perform UNION ALL over
+    * @param graphs              graph to perform UNION ALL over
     * @param preventIdCollisions flag that indicates if ID collisions between entities should be prevented
     * @return union all operator
     */
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
@@ -27,11 +27,9 @@
 package org.opencypher.okapi.relational.api.physical
 
 import org.opencypher.okapi.api.graph.{PropertyGraph, QualifiedGraphName}
-import org.opencypher.okapi.api.schema.Schema
 import org.opencypher.okapi.api.table.CypherRecords
 import org.opencypher.okapi.ir.api.block.SortItem
 import org.opencypher.okapi.ir.api.expr.{Aggregator, Expr, Var}
-import org.opencypher.okapi.ir.api.set.SetPropertyItem
 import org.opencypher.okapi.logical.impl._
 import org.opencypher.okapi.relational.impl.table.{ProjectedExpr, ProjectedField, RecordHeader}
 
@@ -258,7 +256,7 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
   def planCartesianProduct(lhs: P, rhs: P, header: RecordHeader): P
 
   /**
-    * Joins the two input records on given variables.
+    * Joins the two input records using the given expressions.
     *
     * @param lhs         first previous operator
     * @param rhs         second previous operator
@@ -325,30 +323,6 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
   def planExistsSubQuery(lhs: P, rhs: P, targetField: Var, header: RecordHeader): P
 
   // Ternary operators
-
-  /**
-    * Expands the records in the first input (nodes) via the records in the second input (relationships) into the
-    * records in the third input (nodes).
-    *
-    * @param first                   first previous operator
-    * @param second                  second previous operator
-    * @param third                   third previous operator
-    * @param source                  node variable in the first input
-    * @param rel                     relationship variable in the second input
-    * @param target                  node variable in the third input
-    * @param header                  resulting record header
-    * @param removeSelfRelationships set true, iff loops shall be removed from the output
-    * @return expand source operator
-    */
-  def planExpandSource(
-    first: P,
-    second: P,
-    third: P,
-    source: Var,
-    rel: Var,
-    target: Var,
-    header: RecordHeader,
-    removeSelfRelationships: Boolean = false): P
 
   /**
     * Performs a bounded variable length path expression.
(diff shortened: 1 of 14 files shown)


okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
```diff
@@ -258,15 +256,15 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
...
-    * @param header     resulting record header
-    * @return value join operator
+    * @param lhs         first previous operator
+    * @param rhs         second previous operator
+    * @param joinColumns sequence of left and right join columns
+    * @param header      resulting record header
+    * @return join operator
     */
-  def planValueJoin(lhs: P, rhs: P, predicates: Set[org.opencypher.okapi.ir.api.expr.Equals], header: RecordHeader): P
+  def planJoin(lhs: P, rhs: P, joinColumns: Seq[(Expr, Expr)], header: RecordHeader): P
```
Mats-SX: This generalisation is great! But I wonder why it is a sequence and not a set?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
```diff
@@ -315,30 +299,6 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
...
-    * @param second                  second previous operator
-    * @param third                   third previous operator
-    * @param source                  node variable in the first input
-    * @param rel                     relationship variable in the second input
-    * @param target                  node variable in the third input
-    * @param header                  resulting record header
-    * @param removeSelfRelationships set true, iff loops shall be removed from the output
-    * @return expand source operator
-    */
-  def planExpandSource(
```
Mats-SX: Great wins removing these

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
```diff
@@ -141,26 +142,41 @@ class PhysicalPlanner[P <: PhysicalOperator[R, G, C], R <: CypherRecords, G <: P
...
+            producer.planJoin(tempResult, third, Seq(endNode -> target), header)
+
           case Undirected =>
-            val outgoing = producer.planExpandSource(first, second, third, source, rel, target, header)
-            val incoming = producer.planExpandSource(third, second, first, target, rel, source, header, removeSelfRelationships = true)
+            val tempOutgoing = producer.planJoin(first, second, Seq(source -> startNode), first.header ++ second.header)
+            val outgoing = producer.planJoin(tempOutgoing, third, Seq(endNode -> target), header)
+
+            val filterExpression = Not(Equals(startNode, endNode)(CTBoolean))(CTBoolean)
+            val relsWithoutLoops = producer.planFilter(second, filterExpression, second.header)
```
Mats-SX: I know you just moved this one, but now I'm curious as to why it is there...

spark-cypher/src/main/scala/org/opencypher/spark/impl/DataFrameOps.scala
```diff
@@ -198,7 +198,7 @@ object DataFrameOps {
 
       val joinExpr = joinCols.map {
         case (l, r) => df.col(l) === other.col(r)
-      }.foldLeft(lit(true))((acc, expr) => acc && expr)
+      }.reduce((acc, expr) => acc && expr)
```
Mats-SX: `.reduce()` fails on an empty sequence. Are we sure this is never empty here?

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/BinaryOperators.scala
```diff
@@ -45,39 +46,46 @@ private[spark] abstract class BinaryPhysicalOperator extends CAPSPhysicalOperato
...
-    val slots = predicates.map { p =>
-      leftHeader.slotsFor(p.lhs).head -> rightHeader.slotsFor(p.rhs).head
-    }.toSeq
+    implicit context: CAPSRuntimeContext): CAPSPhysicalResult = {
+
+    val joinSlots = joinColumns.map {
+      case (leftExpr, rightExpr) =>
+        val leftRecordSlot = header.slotsFor(leftExpr)
+          .headOption
+          .getOrElse(throw IllegalArgumentException("Expression mapping to a single column", leftExpr))
```
Mats-SX: Nice improvement of the error case.

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/CAPSPhysicalOperator.scala
```diff
@@ -71,9 +71,14 @@ object CAPSPhysicalOperator {
...
 
-    val joinCols = joinSlots.map(pair => columnName(pair._1) -> columnName(pair._2))
-
-    joinDFs(lhsData, rhsData, header, joinCols)(joinType, deduplicate)(lhs.caps)
+    val joinCols = joinSlots.map { case (left, right) => columnName(left) -> columnName(right) }
+
+    // TODO: the join produced corrupt data when the previous operator was a cross. We work around that by using a
+    // subsequent select. This can be removed, once https://issues.apache.org/jira/browse/SPARK-23855 is solved or we
+    // upgrade to Spark 2.3.0
+    val weirdResult = joinDFs(lhsData, rhsData, header, joinCols)(joinType, deduplicate)(lhs.caps)
```
Mats-SX: Perhaps we should rename this variable to something like `potentiallyCorruptedResult`? It isn't guaranteed to be weird.

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/UnaryOperators.scala
```diff
@@ -71,22 +69,28 @@ final case class Cache(in: CAPSPhysicalOperator) extends UnaryPhysicalOperator w
       prev
     })
   }
+}
 
+final case class NodeScan(in: CAPSPhysicalOperator, v: Var, header: RecordHeader) extends UnaryPhysicalOperator {
```
Mats-SX: How come you split this into two different operators? I'm not opposed, but I'm curious for the motivation.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
s1ck: To avoid duplicating self-loops in the undirected case.

spark-cypher/src/main/scala/org/opencypher/spark/impl/DataFrameOps.scala
s1ck: If it's empty, we're doing something wrong. So I guess, it would be good if it fails at this point.

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/UnaryOperators.scala
s1ck: I guess the motivation was to quickly see the difference when looking at physical plans. I would keep it for now.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
s1ck: Didn't see a reason for having a set.

PR #342, opened 2018-04-09
Fix recommendation example

freeclimbing:
The example expected the `CypherSession.store` function to return a qualified graph name. 

Co-Authored-By: Martin Junghanns <martin.junghanns@neotechnology.com>


Mats-SX:
@freeclimbing Hello Markus and thank you for this contribution. In order for us to accept it, you need to sign the openCypher Contributors Agreement (oCCA), for which you find instructions here: http://www.opencypher.org/contributions

In short, you need to send an email to occa@opencypher.org with the attached PDF (see above link) including your full name, email address and organisation (if you're representing one). You also need to amend your commit message to include a sign-off line, with the same email address as your oCCA signature.

Mats-SX:
CLA signature confirmed!

PR #355, opened 2018-04-16
Refactor RecordHeader to Map[Expr, Set[Var]]

pstutz:
Most of the tests break because I'm working on turning `Alias` from a physical operation into adding a header field for the same `Expr`instead.


pstutz:
Obsolete now we have merged the new RH.

PR #357, opened 2018-04-17
Escape quotation marks in Cypher strings

Mats-SX:


spark-cypher/src/test/scala/org/opencypher/spark/api/value/CAPSValueToStringTest.scala
```diff
@@ -24,12 +24,13 @@
...
-package org.opencypher.okapi.api.value
+package org.opencypher.spark.api.value
 
 import org.opencypher.okapi.api.value.CypherValue._
-import org.opencypher.spark.test.CAPSTestSuite
+import org.opencypher.okapi.api.value.{CAPSNode, CAPSRelationship}
+import org.opencypher.okapi.test.BaseTestSuite
 
-class CAPSValueToStringTest extends CAPSTestSuite {
+class CAPSValueToStringTest extends BaseTestSuite {
```
s1ck: Could we move those tests into `okapi-api` or `okapi-ir`, e.g. by using `TestNode` instead of `CAPSNode`?

PR #360, opened 2018-04-19
Dynamically compute printed column widths

### LINKED ISSUE #312: Fix RecordPrinter for entities
Currently the record printer works with fixed column widths. For the most entries this strips the bigger part of the entries output

pstutz:
Fixes #312

### DIFF
--- a/okapi-api/src/main/scala/org/opencypher/okapi/impl/table/RecordsPrinter.scala
+++ b/okapi-api/src/main/scala/org/opencypher/okapi/impl/table/RecordsPrinter.scala
@@ -27,6 +27,7 @@
 package org.opencypher.okapi.impl.table
 
 import org.opencypher.okapi.api.table.CypherRecords
+import org.opencypher.okapi.api.value.CypherValue
 import org.opencypher.okapi.impl.util.PrintOptions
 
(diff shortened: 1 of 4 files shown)


okapi-api/src/main/scala/org/opencypher/okapi/impl/table/RecordsPrinter.scala
```diff
@@ -37,45 +38,65 @@ object RecordsPrinter {
...
+
+    val rows: Seq[CypherValue.CypherMap] = records.collect
+
+    // Dynamically sets the column width for each column to the longest string stored in the column
+    val columnWidths: Map[String, Int] = {
+      for {
+        columnName <- columnNames
+        headerWidth = columnName.length
+        columnRowValueLengths = rows.map(row => row.getOrElse(columnName).toCypherString.length)
+        maxColumnRowValueWidth = columnRowValueLengths.foldLeft(0)(math.max)
```
DarthMax: Instead of 0 you could start with headerWidth

pstutz: Nice find, thanks

PR #361, opened 2018-04-19
Relational optional and unwind

DarthMax:


okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/physical/PhysicalOperatorProducer.scala
```diff
@@ -87,16 +88,37 @@ trait PhysicalOperatorProducer[P <: PhysicalOperator[R, G, C], R <: CypherRecord
...
     * Renames the column identified by the given expression to the specified alias.
     *
     * @param in     previous operator
     * @param expr   expression to be aliased
     * @param alias  alias
     * @param header resulting record header
-    * @return empty records operator
+    * @return Alias operator
+    */
+  def planAlias(in: P, expr: Expr, alias: Var, header: RecordHeader): P = planAlias(in, Seq(expr -> alias), header)
```
s1ck: I don't think we need convenience on that interface. Imho, just keep the upper method.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
```diff
@@ -217,6 +223,56 @@ class PhysicalPlanner[P <: PhysicalOperator[R, G, C], R <: CypherRecords, G <: P
     constructGraphPlan
   }
 
+  private def planOptional(lhs: FlatOperator, rhs: FlatOperator, header: RecordHeader)(implicit context: PhysicalPlannerContext[P, R]) = {
```
s1ck: :heart:

Issue #379, opened 2018-04-24
Feedback 11: Document constraints when turning a DF into an entity table


pstutz:
- Relationships and nodes need IDs, they must be of type `Long`
- If you have multiple node tables as part of same graph, you can’t use overlapping IDs

s1ck:
* columns storing node and relationship identifiers must be named `id`
* columns storing start node identifiers must be named `source`
* columns storing end node identifiers must be named `target`

Mats-SX:
This is now covered by 

https://github.com/opencypher/cypher-for-apache-spark/blob/52f228eb4ee53f0fcdab5eabce075de00e903d0e/okapi-api/src/main/scala/org/opencypher/okapi/api/io/conversion/NodeMapping.scala#L79-L100

and 

https://github.com/opencypher/cypher-for-apache-spark/blob/52f228eb4ee53f0fcdab5eabce075de00e903d0e/okapi-api/src/main/scala/org/opencypher/okapi/api/io/conversion/RelationshipMapping.scala#L146-L158

PR #388, opened 2018-04-24
Simplify entity table construction from DataFrames

### LINKED ISSUE #359: Allow importing a DF with convention instead of an explicit entity mapping
We could use the same convention that already exists for the CSV importer.

Thanks @tomgeudens for the feedback.

s1ck:
Introduces automatic construction of entity mappings by exploiting naming conventions on DataFrame columns. 

Fixes #359

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
@@ -117,11 +117,11 @@ object CAPSNodeTable {
     * columns, the column name is used as property key.
     *
     * @param impliedLabels implied node labels
-    * @param table         Spark table
+    * @param nodeDF        node data
     * @return a node table with inferred node mapping
     */
-  def apply(impliedLabels: Set[String], table: SparkTable): CAPSNodeTable =
-    CAPSNodeTable(impliedLabels, Map.empty, table)
+  def apply(impliedLabels: Set[String], nodeDF: DataFrame): CAPSNodeTable =
+    CAPSNodeTable(impliedLabels, Map.empty, nodeDF)
 
   /**
     * Creates a node table from the given [[DataFrame]]. By convention, there needs to be one column storing node
@@ -131,11 +131,10 @@ object CAPSNodeTable {
     *
     * @param impliedLabels  implied node labels
     * @param optionalLabels mapping from optional labels to column names
-    * @param table          Spark table
+    * @param nodeDF         node data
     * @return a node table with inferred node mapping
     */
-  def apply(impliedLabels: Set[String], optionalLabels: Map[String, String], table: SparkTable): CAPSNodeTable = {
-    val nodeDF = table.df
+  def apply(impliedLabels: Set[String], optionalLabels: Map[String, String], nodeDF: DataFrame): CAPSNodeTable = {
     val nodeProperties = (properties(nodeDF.columns) -- optionalLabels.values).map(col => col -> col).toMap
     val nodeMapping = NodeMapping(GraphEntity.sourceIdKey, impliedLabels, optionalLabels, nodeProperties)
     CAPSNodeTable(nodeMapping, nodeDF)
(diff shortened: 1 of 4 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
```diff
@@ -111,6 +111,36 @@ object CAPSNodeTable {
...
+  /**
+    * Creates a node table from the given [[DataFrame]]. By convention, there needs to be one column storing node
+    * identifiers and named after [[GraphEntity.sourceIdKey]]. All remaining columns are interpreted as node property
+    * columns, the column name is used as property key.
+    *
+    * @param impliedLabels implied node labels
+    * @param table         Spark table
+    * @return a node table with inferred node mapping
+    */
+  def apply(impliedLabels: Set[String], table: SparkTable): CAPSNodeTable =
```
pstutz: We should probably take a DF parameter instead of the implicit SparkTable class. That would make it easier to understand what the parameter is and it would require one less import for the user to get the implicit class.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
```diff
@@ -111,6 +111,36 @@ object CAPSNodeTable {
...
+    * identifiers and named after [[GraphEntity.sourceIdKey]]. Optional labels are defined by a mapping from label to
+    * column name. All remaining columns are interpreted as node property columns, the column name is used as property
+    * key.
+    *
+    * @param impliedLabels  implied node labels
+    * @param optionalLabels mapping from optional labels to column names
+    * @param table          Spark table
+    * @return a node table with inferred node mapping
+    */
+  def apply(impliedLabels: Set[String], optionalLabels: Map[String, String], table: SparkTable): CAPSNodeTable = {
```
pstutz: Same as line 123.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
```diff
@@ -123,7 +153,31 @@ object CAPSRelationshipTable {
...
+      Relationship.sourceStartNodeKey,
+      Relationship.sourceEndNodeKey,
+      relationshipType,
+      relationshipProperties)
+
+    CAPSRelationshipTable(relationshipMapping, relationshipDF)
+  }
+
+  /**
+    * Creates a relationship table from the given [[DataFrame]]. By convention, there needs to be one column storing
```
pstutz: It would be easier for the user if we committed to these key names forever and put the actual names into the documentation. I'm not certain though.

spark-cypher/src/test/scala/org/opencypher/spark/impl/table/EntityTableTest.scala
```diff
@@ -159,4 +159,45 @@ class EntityTableTest extends CAPSTestSuite {
...
     }
   }
+
+  test("NodeTable should infer the correct node mapping") {
+    val df = session.createDataFrame(Seq((1, "Alice", 1984, true, 13.37)))
+      .toDF("id", "name", "birthYear", "isGood", "luckyNumber")
+
+    val nodeTable = CAPSNodeTable(Set("Person"), df)
+
+    nodeTable.schema should equal(Schema.empty
```
pstutz: Maybe add a test for optional labels.

pstutz:
We should also document these conventions when we address https://github.com/opencypher/cypher-for-apache-spark/issues/379

spark-cypher/src/test/scala/org/opencypher/spark/impl/table/EntityTableTest.scala
s1ck: It's below in line 178

pstutz: Sorry, overlooked it.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/EntityTable.scala
s1ck: It would most certainly make the documentation clearer. I don't mind using fixed names. We are still a bit inconsistent yet, e.g.`sourceStartNodeKey` is represented by `source` (instead of `start`).

pstutz: Makes sense, let's make it consistent and then never change it. Should I create a new issue or do we do that when resolving the documentation issue?

s1ck: I added a comment to #379.

spark-cypher/src/test/scala/org/opencypher/spark/impl/table/EntityTableTest.scala
s1ck: No problemo :)

Issue #401, opened 2018-04-26
Add tests for storing/loading graph tags to PGDSAcceptance


pstutz:

pstutz:
Fixed by https://github.com/opencypher/cypher-for-apache-spark/pull/403

PR #403, opened 2018-04-26
Extend PGDS acceptance tests

pstutz:
Add test that requires graph tag support in a CAPS PGDS

Addresses #401


okapi-testing/src/main/scala/org/opencypher/okapi/testing/PGDSAcceptance.scala
```diff
@@ -181,6 +181,58 @@ trait PGDSAcceptance extends BeforeAndAfterAll {
     }
   }
 
+  // TODO: Requires fixing https://github.com/opencypher/cypher-for-apache-spark/issues/402
+  //  it("supports UNION ALl (requires storing/loading graph tags for CAPS)") {
```
s1ck: uncomment and change to `ignore`?

PR #414, opened 2018-04-30
Refactor column selection logic in SelectFields to Physical planning

s1ck:
* fixes a bug in `ColumnName.of(expr: Expr)`
* rename `SelectFields` to `Select`

### DIFF
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/ColumnName.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/ColumnName.scala
@@ -45,20 +45,16 @@ object ColumnName {
       case ProjectedExpr(p: Property) => new NameBuilder() += None += p.withoutType + p.cypherType.material.name
       case ProjectedExpr(expr) => new NameBuilder() += None += expr.withoutType
     }
-
     builder.result()
   }
 
   def of(expr: Expr): String = {
-    expr match {
-      case Var(name) => name
-      case _: Property =>
-        val columnNameBuilder = new NameBuilder() += None += expr.withoutType + expr.cypherType.material.name
-        columnNameBuilder.result()
-      case _ =>
-        val columnNameBuilder = new NameBuilder() += None += expr.withoutType
-        columnNameBuilder.result()
+    val builder = expr match {
+      case Var(name) => new NameBuilder() += name
+      case _: Property => new NameBuilder() += None += expr.withoutType + expr.cypherType.material.name
+      case _ => new NameBuilder() += None += expr.withoutType
     }
+    builder.result()
   }
 
   def from(name: String): String = (new NameBuilder() += name).result()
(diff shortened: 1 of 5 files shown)


okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/ColumnName.scala
```diff
@@ -45,20 +45,16 @@ object ColumnName {
...
   def of(expr: Expr): String = {
-    expr match {
-      case Var(name) => name
-      case _: Property =>
-        val columnNameBuilder = new NameBuilder() += None += expr.withoutType + expr.cypherType.material.name
-        columnNameBuilder.result()
-      case _ =>
-        val columnNameBuilder = new NameBuilder() += None += expr.withoutType
-        columnNameBuilder.result()
+    val builder = expr match {
```
DarthMax: :+1:

spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/UnaryOperators.scala
```diff
@@ -206,31 +213,16 @@ final case class ReturnGraph(in: CAPSPhysicalOperator) extends UnaryPhysicalOper
...
-        }
-      }
 
       val data = records.data
-      val columns = groupedSlots.map { s =>
-        data.col(ColumnName.of(s))
-      }
-      val newData = records.data.select(columns: _*)
+
+      val columns = expressions.map(expr => data.col(ColumnName.of(expr)))
```
DarthMax: Do we ensure that every of the expression in the list actually exist?

s1ck: Not really and I don't understand this to be a job for the back-end implementer. Or asked differently: What would you do if the column does not exist?

Mats-SX: I guess that is enforced much earlier, in the frontend, where nonsensical expressions (like `MATCH (foo) WITH bar.prop`) are caught.

DarthMax: Mhh it is just that in the relational sense select can also evaluate expression as in 
`SELECT 1+2` which one could think would be possible here as well.

s1ck: You're right. But `SELECT 1+2` is actually a projection. To make it more clear, I was thinking about given the select operator just the `String` column names. However, this is also not a good idea, since e.g. `mem-cypher` and `flink-cypher` do not user `ColumnName.of` from CAPS.

Mats-SX: This is the very reason why it was `Set[Var]`; to make it clear it is about selection and not projection.

s1ck: I'll experiment a bit on fusing `Select`, `Project` and `Alias`. I don't see why `Select` should not evaluate expressions, similar to `DataFrame.select`. We could also offer `.withColumn` and `.withColumnRenamed` for covenience.

I think we can merge that PR for now as it takes complexity away from the backend implementers.

PR #416, opened 2018-04-30
Test for correct tags at entities

### LINKED ISSUE #415: CAPSPGDSAcceptance tag testing should be more robust
When the identifiers of a graph source are changed on `.store()` it's not guaranteed that the tags are still in place. The test should allow for this to happen, by checking the actual tag sector of the identifiers and asserting that they correspond to what the `CAPSGraph.tags()` field reports.

s1ck:
* fixes #415

Co-Authored-By: Philip Stutz <philip.stutz@gmail.com>

### DIFF
--- a/spark-cypher/src/test/scala/org/opencypher/spark/api/io/CAPSPGDSAcceptance.scala
+++ b/spark-cypher/src/test/scala/org/opencypher/spark/api/io/CAPSPGDSAcceptance.scala
@@ -94,5 +94,7 @@ trait CAPSPGDSAcceptance extends PGDSAcceptance[CAPSSession] {
 
     nodeTags.foreach(tag => graph.tags should contain(tag))
     relTags.foreach(tag => graph.tags should contain(tag))
+
+    graph.tags -- (nodeTags ++ relTags) shouldBe empty
   }
 }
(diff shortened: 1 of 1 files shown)


spark-cypher/src/test/scala/org/opencypher/spark/api/io/CAPSPGDSAcceptance.scala
```diff
@@ -65,15 +71,28 @@ trait CAPSPGDSAcceptance extends PGDSAcceptance[CAPSSession] {
...
+    val rels = graph.relationships("r").asDataset.map(row => row("r").cast[CAPSRelationship]).collect()
 
+    nodes.length shouldBe expectedNodeSize
+    rels.length shouldBe expectedRelSize
+
+    val nodeTags = nodes.map(_.id.getTag).toSet
+    val relTags = rels.map(_.id.getTag).toSet
+
+    nodeTags.foreach(tag => graph.tags should contain(tag))
+    relTags.foreach(tag => graph.tags should contain(tag))
```
Mats-SX: Beautiful! Should we additionally cross-check that graph.tags doesn't contain any tag that is not in the data?

PR #418, opened 2018-04-30
More error output to pinpoint the cause of our spurious TravisCI failures

pstutz:
Example: Current master build is spuriously failing, same with this PR: https://github.com/opencypher/cypher-for-apache-spark/pull/416


pstutz:
Did not work out, considering dropping TravisCI.

PR #436, opened 2018-05-16
Cypher type name parser

DarthMax:
Adds `CypherType.fromName("INTEGER") -> Some(CTInteger)`

### DIFF
--- a/okapi-api/src/main/scala/org/opencypher/okapi/api/types/CypherType.scala
+++ b/okapi-api/src/main/scala/org/opencypher/okapi/api/types/CypherType.scala
@@ -34,6 +34,58 @@ import scala.language.postfixOps
 
 object CypherType {
 
+  /**
+    * Parses the name of CypherType into the actual CypherType object.
+    *
+    * @param name string representation of the CypherType
+    * @return
+    * @see {{{org.opencypher.okapi.api.types.CypherType#name}}}
+    */
+  def fromName(name: String): Option[CypherType] = {
+    def extractLabels(s: String, typ: String, sep: String): Set[String] = {
+      val regex = s"""$typ\\(:(.+)\\)\\??""".r
+      s match {
+        case regex(l) => l.split(sep).toSet
+        case _ => Set.empty[String]
+      }
+    }
+
+    val graphRegex = """\@ (.+)\$""".r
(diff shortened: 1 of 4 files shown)


okapi-api/src/main/scala/org/opencypher/okapi/api/types/CypherType.scala
```diff
@@ -34,6 +34,58 @@ import scala.language.postfixOps
...
+      case "PATH"    | "PATH?"    => Some(CTPath)
+      case "?"       | "??"       => Some(CTWildcard)
+
+      case node if node.startsWith("NODE") =>
+        Some(CTNode(extractLabels(node, "NODE", ":"), extractGraph(node)))
+
+      case rel if rel.startsWith("RELATIONSHIP") =>
+        Some(CTRelationship(extractLabels(rel, "RELATIONSHIP", """\|"""), extractGraph(rel)))
+
+      case list if list.startsWith("LIST") =>
```
pstutz: Does this approach work for nested structures?

okapi-api/src/test/scala/org/opencypher/okapi/api/types/CypherTypesTest.scala
```diff
@@ -115,6 +115,12 @@ class CypherTypesTest extends FunSpec with Matchers {
     CTNull.toString shouldBe "NULL"
   }
 
+  it("can parse CypherType names into CypherTypes"){
+    allTypes.foreach { t =>
+      CypherType.fromName(t.name).get should equal(t)
```
pstutz: Did you test the graph extraction?

pstutz:
Thanks a lot! 👍 I think this might be useful elsewhere and it's not easy to get right.

okapi-api/src/main/scala/org/opencypher/okapi/api/types/CypherType.scala
DarthMax: There are tests with Lists in the test suite. I think it should work

okapi-api/src/test/scala/org/opencypher/okapi/api/types/CypherTypesTest.scala
DarthMax: good point

PR #444, opened 2018-05-19
Add data source for reading graphs from edge lists

s1ck:
@pstutz would be nice if you could have a look. Maybe we can integrate it better with the new data source abstraction. I tried using `AbstractDataSource` but the overhead was too large for such a simple (read only) implementation.


pstutz:
I think it's fine if this is its own thing. Abstract data source mainly does caching when there are several graphs and translation of store/load operations, and this one can only load a single graph. Maybe there's something reusable to extract for read-only DS? We could figure that out once we have more of them.

PR #458, opened 2018-05-24
Make WKDS internals more accessible

### LINKED ISSUE #454: Expose relative paths within the WKDS for file based graph sources

pstutz:
Fixes #454

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
@@ -56,7 +56,19 @@ object DefaultGraphDirectoryStructure {
     def path: String = graphName.value.replace(".", pathSeparator)
   }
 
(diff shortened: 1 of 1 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
```diff
@@ -56,7 +56,19 @@ object DefaultGraphDirectoryStructure {
     def path: String = graphName.value.replace(".", pathSeparator)
   }
 
-  val pathSeparator = Path.SEPARATOR
+  val pathSeparator: String = Path.SEPARATOR
```
s1ck: Should we maybe move those as `def`s to the trait and let the `DefaultGraphDirectoryStructure` forward the calls to the object?

pstutz: Good idea, created an issue for it #462

PR #461, opened 2018-05-24
Allow custom SparkSession init

### LINKED ISSUE #459: Would like to create a custom Spark session (e.g. enableHiveSupport) for the CAPS session.
Would like to do something like the following:
```scala
 override def initSession(): CAPSSession = {
    implicit val session: SparkSession = SparkGateway(true)
    CAPSSession.create(session)
  }
```
The above doesn't work - I suspect this is something to do with a `caps` session already having an `implicit` in `CAPSSessionFixture` (see below).  This will only work if I return the `caps` instance for in the above `initSession`

```scala
trait CAPSSessionFixture extends BaseTestFixture {
  self: SparkSessionFixture with BaseTestSuite =>

  implicit lazy val caps: CAPSSession = CAPSSession.create

  abstract override protected def afterEach(): Unit = {
    // delete all session graphs via their qualified graph name
    caps.catalog.source(caps.catalog.sessionNamespace).graphNames.map(_.value).foreach(caps.catalog.delete)
    caps.catalog.store(caps.emptyGraphQgn, CAPSGraph.empty)
    super.afterEach()
  }
}
```

pstutz:
Fixes #459

### DIFF
--- a/spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/fixture/SparkSessionFixture.scala
+++ b/spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/fixture/SparkSessionFixture.scala
@@ -33,5 +33,8 @@ import org.opencypher.spark.testing.TestSparkSession
 trait SparkSessionFixture extends BaseTestFixture {
   self: BaseTestSuite =>
 
-  implicit val session: SparkSession = TestSparkSession.instance
+  implicit lazy val sparkSession: SparkSession = initSparkSession
+
+  def initSparkSession: SparkSession = TestSparkSession.instance
+
 }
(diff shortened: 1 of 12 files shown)


spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/fixture/SparkSessionFixture.scala
```diff
@@ -33,5 +33,8 @@ import org.opencypher.spark.testing.TestSparkSession
 trait SparkSessionFixture extends BaseTestFixture {
   self: BaseTestSuite =>
 
-  implicit val session: SparkSession = TestSparkSession.instance
+  implicit lazy val sparkSession: SparkSession = initSparkSession
+
+  def initSparkSession: SparkSession = TestSparkSession.instance
```
s1ck: Was looking for the actual change :) found it! :+1:

Issue #478, opened 2018-05-27
Investigate flaky builds due to StackoverflowException in Typer


s1ck:
builds fail quite often due to 
```
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typed1(Typers.scala:5387)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.transform.Erasure$Eraser.typed1(Erasure.scala:712)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.runTyper$1(Typers.scala:5423)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.scala$tools$nsc$typechecker$Typers$Typer$$typedInternal(Typers.scala:5450)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.body$2(Typers.scala:5397)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typed(Typers.scala:5401)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typed(Typers.scala:5492)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typedLabelDef(Typers.scala:2291)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typedOutsidePatternMode$1(Typers.scala:5361)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typedInAnyMode$1(Typers.scala:5380)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.typed1(Typers.scala:5387)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.transform.Erasure$Eraser.typed1(Erasure.scala:712)
[22:17:49][org.opencypher:spark-cypher] 	at scala.tools.nsc.typechecker.Typers$Typer.runTyper$1(Typers.scala:5423)
[...]

PR #479, opened 2018-05-27
Investigate StackoverflowError in Typer

s1ck:
* Set -Ytyper-debug in maven-scala-plugin
* #478


pstutz:
Managed to reproduce it locally. I think it's related to shapeless magic used by circe serialisation.

s1ck:
Yeah, I was thinking the same since it's the only magic happening in that PR. Did we configure it wrong?

PR #484, opened 2018-05-28
Fix nullability conversion for nested types

s1ck:
Co-Authored-By: Philip Stutz <philip.stutz@gmail.com>

### DIFF
--- a/okapi-api/src/test/scala/org/opencypher/okapi/api/types/CypherTypesTest.scala
+++ b/okapi-api/src/test/scala/org/opencypher/okapi/api/types/CypherTypesTest.scala
@@ -51,6 +51,7 @@ class CypherTypesTest extends FunSpec with Matchers {
     CTList(CTAny),
     CTList(CTList(CTBoolean)),
     CTList(CTWildcard),
+    CTList(CTString.nullable),
     CTWildcard,
     CTVoid
   )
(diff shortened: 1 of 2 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/impl/DataFrameOps.scala
```diff
@@ -145,6 +145,35 @@ object DataFrameOps {
       }
     }
 
+    def setNullability(mapping: Map[String, CypherType]): DataFrame = {
+      val newSchema = StructType(df.schema.map {
+
+        case s@StructField(cn, array: ArrayType, _, _) if mapping.contains(cn) =>
```
pstutz: Could we have an implicit class that offers `CypherType.toStructField` that handles nested types recursively? It might allow us to use:
```
val newSchema = df.schema.map {
  case s@StructField(cn, _, _, _) => mapping(cn).toStructField
  case other => throw InvalidArgumentException(s"No mapping provided for $other")
}
```

pstutz: I think that implicit class already exists as `CAPSRecordHeader#CAPSRecordSlot`

s1ck: Good idea, added! It's a bit more involved then just checking nullability, but it's better separated and reusable.

s1ck: Oh I didn't see your second comment. Should we align those two? Maybe let the RecordHeader one call the CAPSCypherType implicits?

pstutz: Yes, let's align them, not even sure if the other one handled nullability correctly.

PR #497, opened 2018-06-05
Record header refactoring

s1ck:


spark-cypher/src/main/scala/org/opencypher/spark/impl/operations.scala
```diff
@@ -1,52 +0,0 @@
...
- * proposals for change that have been documented or implemented should only be
- * described as "implementation extensions to Cypher" or as "proposed changes to
- * Cypher that are not yet approved by the openCypher community".
- */
-package org.opencypher.spark.impl
-
-import org.opencypher.okapi.api.value.CypherValue.CypherMap
-import org.opencypher.okapi.ir.api.expr.{Expr, Var}
-
-object operations {
```
Mats-SX: I know this isn't really being used, but is there a particular reason for removing them?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
```diff
@@ -0,0 +1,246 @@
...
+  def from(header: RecordHeader, table: T, displayNames: Option[Seq[String]] = None): R
+
+  def table: T
+
+  override def physicalColumns: Seq[String] = table.physicalColumns
+
+  def header: RecordHeader
+
+  def select(expr: (Expr, Option[Var]), epxrs: (Expr, Option[Var])*): R = {
+    val allExprs = expr +: epxrs
```
pstutz: `epxrs`

PR #499, opened 2018-06-07
Fix and test CypherValue.toCypherString

DarthMax:
Fixes https://github.com/neo-technology/morpheus/issues/98


okapi-api/src/test/scala/org/opencypher/okapi/api/value/CypherValueTest.scala
```diff
@@ -0,0 +1,84 @@
...
+  case class TestNode(
+    override val id: Long,
+    override val labels: Set[String] = Set.empty,
+    override val properties: CypherMap = CypherMap.empty) extends CypherNode[Long] {
+
+    override type I = TestNode
+
+    override def copy(id: Long = id, labels: Set[String] = labels, properties: CypherMap = properties): TestNode = {
+      TestNode(id, labels, properties)
+    }
```
Mats-SX: We have `CAPSValueToStringTest` as well, just fyi

PR #503, opened 2018-06-10
Replace Circe with uJson, add `CypherValue.toJson`

### LINKED ISSUE #485: Replace circe with uPickle in Zeppelin JsonSerialization

pstutz:
Fixes #485


pstutz:
@DarthMax Does Zeppelin network visualisation support String IDs? Couldn't get a definitive answer in the documentation. Converting a `Long` to a JavaScript `Number` is not safe: [`JavaScript has a Number type which is a 64 bit floating point number`](https://stackoverflow.com/questions/5450012/how-to-convert-a-string-to-long-in-javascript?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa).

spark-cypher-testing/src/test/scala/org/opencypher/spark/web/CAPSJsonSerialiserTest.scala
```diff
@@ -1,451 +0,0 @@
...
-import org.opencypher.okapi.ir.api.expr.Var
-import org.opencypher.okapi.relational.impl.syntax.RecordHeaderSyntax._
-import org.opencypher.okapi.relational.impl.table.{OpaqueField, RecordHeader}
-import org.opencypher.spark.impl.{CAPSGraph, CAPSRecords}
-import org.opencypher.spark.testing.CAPSTestSuite
-import org.opencypher.spark.testing.fixture.TeamDataFixture
-import org.opencypher.spark.web.CAPSJsonSerialiser.toJsonString
-
-//noinspection NameBooleanParameters
-class CAPSJsonSerialiserTest extends CAPSTestSuite with TeamDataFixture {
```
s1ck: phew ... luckily it took me only ~30 seconds to fix those test in the RH PR :)

okapi-api/src/main/scala/org/opencypher/okapi/api/value/CypherValue.scala
```diff
@@ -181,6 +185,33 @@ object CypherValue {
       }
     }
 
+    def toJson: Js.Value = {
```
s1ck: nice!

DarthMax:
Mhh I have no idea. For the visualization case it would not really have an implication if the reverse operation of to json is not safe

pstutz:
We should probably have a more efficient Json table representation. Array of header names, then rows that are arrays.

DarthMax:
I think it was used for the simple web service we had build as a show case once. 
I think we can keep it as it is. The strength of JSON is readability and accessibility not memory efficient storage.

pstutz:
Removed `CypherRecords.toJson` after additional conversation with @DarthMax

PR #506, opened 2018-06-11
Remove all remaining hard-coded column accesses (preparatory for DF caching)

### LINKED ISSUE #424: Remove variable name from node/rel scans on our main PropertyGraph interfaces
Requires unnecessary renaming and prevents caching of node/rel scans.

pstutz:


spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
```diff
@@ -203,8 +203,8 @@ case class CAPSRecords(
     }
 
     // add missing columns
-    val (dataWithMissingColumns, _) = missingExpressions.foldLeft(dataWithColumnsRenamed -> updatedHeader) {
-      case ((currentDf, currentHeader), expr) =>
+    val dataWithMissingColumns = missingExpressions.foldLeft(dataWithColumnsRenamed) {
```
Mats-SX: Nice simplification!

PR #509, opened 2018-06-13
Generalize and simplify printing of tabular data

### LINKED ISSUE #508: Generalize and simplify RecordsPrinter

s1ck:
prints tables like the following:
```
╔═════════════════╤═════════╗
║ product         │ name    ║
╠═════════════════╪═════════╣
║ 'Terminator 2'  │ 'Eve'   ║
║ 'Cryptonomicon' │ 'Trudy' ║
║ '1984'          │ 'Bob'   ║
╚═════════════════╧═════════╝
(3 rows)
```
* fixes #508


okapi-api/src/main/scala/org/opencypher/okapi/api/value/CypherValue.scala
```diff
@@ -86,6 +86,8 @@ object CypherValue {
     Option(cv).flatMap(v => Option(v.value))
   }
 
+  def toCypherString(value: CypherValue): String = value.toCypherString
```
DarthMax: Seems like a strange duplication

okapi-api/src/main/scala/org/opencypher/okapi/impl/table/RecordsPrinter.scala
```diff
@@ -40,87 +41,14 @@ object RecordsPrinter {
...
-  }
-
-  private def repeat(x: String, size: Int): String = (1 to size).map((_) => x).mkString
-
-  private def fitToColumn(s: String, width: Int)(implicit options: PrintOptions) = {
-    val spaces: String = (1 until width).map(_ => " ").foldLeft("")(_ + _)
-    val cell = s + spaces
-    cell.take(width)
+    options.stream
+      .append(TablePrinter.toTable(columns, rows)(CypherValue.toCypherString))
```
DarthMax: maybe replace with
`.append(TablePrinter.toTable(columns, rows)( v => v.toCypherString))`
That's not less readable and would make `CypherValue.toCypherString` obsolete again

okapi-api/src/main/scala/org/opencypher/okapi/api/value/CypherValue.scala
s1ck: You're right. I changed it. However, I was wondering why this method has to live at the instance?

PR #511, opened 2018-06-13
Relational Var Expand

DarthMax:
Known regressions:
- binding the var length relationship and returning it as List of Relationship
- undirected var length expand


okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
```diff
@@ -275,6 +275,19 @@ case class RecordHeader(exprToColumn: Map[Expr, String]) {
...
 
+  def join(other: RecordHeader): RecordHeader = {
+    val expressionOverlap = expressions.intersect(other.expressions)
+    if(expressionOverlap.nonEmpty) {
+      throw IllegalArgumentException("two headers with non overlapping expressions", s"overlapping expressions: $expressionOverlap")
+    }
+
+    val cleanOther = if (columns.intersect(other.columns).nonEmpty) {
+      RecordHeader.empty.withExprs(other.expressions)
+    } else other
```
s1ck: Two thoughts about that logic:

1. What happens if there is a n:1 relation between expressions and columns in the right header? Wouldn't we introduce unnecessary columns later?
2. An optimization would be to get the expressions for the conflicting columns and only reinsert those (after adding the remaining the remaining expressions with preserved column names)

PR #514, opened 2018-06-14
tableStorageFormat param not needed for methods on FileBasedDatSource…

hannesmiller:
… - already defined on constructor


pstutz:
@hannesmiller Before the change I think a DS always used the JSON-configured format for reading graphs and the parameter-configured format for writing graphs. With this change the JSON-configured format is useless. I suggest that we either remove it or add a validation that ensures that the formats match.

PR #515, opened 2018-06-14
Support Undirected Var Length Expand

DarthMax:


okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
```diff
@@ -176,12 +175,21 @@ I <: RuntimeContext[A, P]](producer: PhysicalOperatorProducer[O, K, A, P, I])
...
         header, isExpandInto
       ) =>
-        planBoundedVarLengthExpand(
-          source, edgeScan, innerNode, target,
-          direction, lower, upper,
-          sourceOp, edgeScanOp, innerNodeOp, targetOp,
-          header, isExpandInto
-        )
+        val planner = direction match {
+          case Directed => new DirectedVarLengthExpandPlanner[O, K, A, P, I](
```
Mats-SX: I just love how this reads xD

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/VarLengthExpandPlanner.scala
```diff
@@ -0,0 +1,353 @@
...
+    alignedOps.reduce(producer.planTabularUnionAll)
+  }
+
+  /**
+    * Creates the isomorphism filter for the given edge list
+    *
+    * @param rel        new edge
+    * @param candidates candidate edges
+    */
+  protected def isomorphismFilter(rel: Var, candidates: Set[Var]): Ands = Ands(
```
Mats-SX: I like how `isomorphismFilter` is extracted to its own thing here. This implies pluggability for different morphisms.

It should perhaps be `morphismFilter` if we want a general name, or `cypherMorphismFilter` or `edgeIsomorphismFilter` if we want a specific name. We don't do node isomorphism, which I guess is the default interpretation when one just reads 'isomorphism'.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/VarLengthExpandPlanner.scala
```diff
@@ -0,0 +1,353 @@
...
+        val (outIn, _) = expand(i, last, reversedExpandCacheOp, Outbound, edgeVars)
+        val (inOut, _) = expand(i, lastRevered, expandCacheOp, Inbound, edgeVars)
+        val (inIn, _) = expand(i, lastRevered, reversedExpandCacheOp, Inbound, edgeVars)
+        val nextOps = producer.planTabularUnionAll(outOut, inOut) -> producer.planTabularUnionAll(outIn, inIn)
+
+        acc :+ nextOps -> (edgeVars :+ nextEdge)
+    }.filter(_._2.size >= lower)
+
+
+    // Join target nodes on expand ops
```
Mats-SX: Shouldn't this comment say `union`?

PR #516, opened 2018-06-15
Feature/parallelise store

hannesmiller:


spark-cypher/src/main/scala/org/opencypher/spark/api/io/AbstractDataSource.scala
```diff
@@ -135,13 +141,28 @@ abstract class AbstractDataSource(implicit val session: CAPSSession) extends CAP
     writeCAPSGraphMetaData(graphName, CAPSGraphMetaData(tableStorageFormat, capsGraph.tags))
     writeSchema(graphName, schema)
 
-    schema.labelCombinations.combos.foreach { combo =>
-      writeNodeTable(graphName, combo, capsGraph.canonicalNodeTable(combo))
+    val nodeWrites = schema.labelCombinations.combos.map { combo =>
+      Future {
+        writeNodeTable(graphName, combo, capsGraph.canonicalNodeTable(combo))
```
Mats-SX: Is this going to use several threads on each worker node, meaning it will write concurrently on each node, in addition to writing in parallel between the worked machines?

s1ck: @hannesmiller any updates on this?

Mats-SX:
I did a small local test where I found that this was significantly faster than the status quo. I think we might benefit from these changes.

Mats-SX:
Sure; it's quite a simple test. Just push a graph through `.store` with and without these changes and see how it compares.

DarthMax:
I tested it on Mattis with SF10 conversion from Neo4j to Parquet.
Without this PR this takes 23 minutes, with this PR it's 9 minutes. Of those 9 2 minutes are schema procedure in Neo4j

PR #517, opened 2018-06-16
Continued refactoring on physical layer

s1ck:


s1ck:
@DarthMax let's investigate together how we can work around the aliasing issue in var expand.

s1ck:
@pstutz @DarthMax using the `AliasExpr` I was wondering if we could remove the `.withAlias` calls from record header and handle only expressions and react according to the input. wdyt?

s1ck:
@DarthMax I replaced your `expr as expr` logic with `copyColumn` and `renameColumn`. I hope that covers all corner cases.

pstutz:
@s1ck sound reasonable. Just need a good answer to what happens if the aliased thing is removed. Maintaining the previous semantics would probably be tricky:
- We can't have an alias that points at something that's no longer there
- We can't leave the original around after its removal
- It would be unexpected if the previous alias expression can no longer be resolved.

I guess the easiest solution is to try to first resolve an alias as the inner expression itself and only resolve the original expression if the alias expression is not found directly in the header. 
Maybe you have a better idea.

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/ExpandIntoBehaviour.scala
```diff
@@ -108,7 +109,7 @@ class ExpandIntoBehaviour extends CAPSTestSuite with DefaultGraphInit {
     ))
   }
 
-  test("Expand into after var expand") {
+  ignore("Expand into after var expand") {
```
s1ck: Due to the new abstraction, we now need three to four abstract operators to achieve the same behavior in var expand. This leads to additional sql operators and extreme planning times. We should introduce convenience ops where necessary.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/VarLengthExpandPlanner.scala
```diff
@@ -124,30 +124,44 @@ I <: RuntimeContext[A, P]] {
...
-      .select(nextEdge, nextNode)
 
     val aliasedCacheOp = producer.planAliases(
       expandCacheOp, Seq(edgeScan as nextEdge, innerNode as nextNode),
       aliasedCacheHeader
     )
 
+    // We just want to select id columns, select always selects also the children. we need to drop them first
+    // TODO: this is a planning performance killer, we need to squash these steps into a single table operation
+    val idExprs = Set(
```
s1ck: planAliases now forwards to `RelationalCypherRecords#withAliases`. This requires us to perform additional operators to get the relevant data.

Here we are only interested in the id expressions of `nextEdge` and `nextNode`. However, if we pass them into select, their children are selected to. This is why drop the children first before selecting the id columns. Imho, not being able to address the id expression on it's own is a disadvantage here.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/PhysicalPlanner.scala
```diff
@@ -248,11 +248,13 @@ I <: RuntimeContext[A, P]](val producer: PhysicalOperatorProducer[O, K, A, P, I]
 
     // 3. Rename the join expressions on the right hand side, in order to make them distinguishable after the join
     val joinExprRenames = joinExprs.map(e => e as Var(generateUniqueName)(e.cypherType))
-    val rhsHeaderWithRenames = rhsHeaderWithDropped.withAlias(joinExprRenames.toSeq: _*) -- joinExprs
-    val rhsWithRenamed = producer.planAliases(rhsWithDropped, joinExprRenames.toSeq, rhsHeaderWithRenames)
+    val rhsHeaderWithRenames = rhsHeaderWithDropped.withAlias(joinExprRenames.toSeq: _*)
+    val rhsWithAlias = producer.planAliases(rhsWithDropped, joinExprRenames.toSeq, rhsHeaderWithRenames)
+    val rhsHeaderJoinReady = rhsHeaderWithRenames -- joinExprs
+    val rhsJoinReady = producer.planDrop(rhsWithAlias, joinExprs.asInstanceOf[Set[Expr]], rhsHeaderJoinReady)
```
pstutz: Could we pattern match or have a more constrained type instead of using `asInstanceOf`?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/VarLengthExpandPlanner.scala
pstutz: @DarthMax told me about the path itself only being an expression embedded in the segment expressions in the header, but not having its its own entry and column. Maybe a variable should be the same thing. In that case the ID column could explicitly be the id column for that variable (which has no explicit header expression), and we could still select the id column without implicitly selecting everything that is part of the variable.

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/ExpandIntoBehaviour.scala
pstutz: Sounds good, let's add more powerful abstractions that efficiently package up multiple smaller operations.

In theory we could also rewrite many small inefficient ops into more efficient ones. That would be the superior approach wherever the more powerful abstraction wouldn't contribute to code readability and would only exist for performance reasons.

okapi-relational/src/test/scala/org/opencypher/okapi/relational/impl/table/RecordHeaderTest.scala
```diff
@@ -27,7 +27,6 @@
 package org.opencypher.okapi.relational.impl.table
```
pstutz: What happens to an alias when the aliased expression is removed from the header? Could we add a test for that?

pstutz:
GitHub is being crazy and hid my comments. :(

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/physical/VarLengthExpandPlanner.scala
s1ck: Sounds like a nicer approach.

PR #518, opened 2018-06-18
Basic path support

DarthMax:
Lays the basis for path support in Expressions and RecordHeader

TODO:

- [ ] write tests for RecordHeader changes


okapi-api/src/main/scala/org/opencypher/okapi/api/value/CypherValue.scala
```diff
@@ -284,6 +285,23 @@ object CypherValue {
...
   }
 
+  trait CypherPath[Id] extends Product with MaterialCypherValue[Seq[CypherEntity[Id]]] {
+    type I <: Seq[CypherEntity[Id]]
+
+    def value: Seq[CypherEntity[Id]]
+
+    override def unwrap: Any = value.map(_.unwrap)
+
+    override def toString: String = s"${getClass.getSimpleName}(${value.mkString("--")})"
```
pstutz: Could you delegate to `toCypherString`?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -59,16 +59,16 @@ sealed abstract class Expr extends AbstractTreeNode[Expr] {
     * Returns the node/relationship that this expression is owned by, if it is owned.
     * A node/relationship owns its label/key/property mappings
     */
-  def owner: Option[Var] = None
+  def owner(): Option[EntityExpr] = None
```
pstutz: Parentheses necessary?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +79,34 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
   override def withoutType: String = s"$$$name"
 }
 
-final case class Var(name: String)(val cypherType: CypherType = CTWildcard) extends Expr {
+trait EntityExpr extends Expr {
```
s1ck: `alias` is not necessarily an `EntityExpr`. Maybe `EntityExpr` should extend `Var` instead?

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOptimizer.scala
```diff
@@ -62,10 +62,16 @@ object LogicalOptimizer extends DirectCompilationStage[LogicalOperator, LogicalO
   }
 
   def discardScansForNonexistentLabels: PartialFunction[LogicalOperator, LogicalOperator] = {
-    case scan@NodeScan(v, in, _) =>
+    case scan@NodeScan(eexpr, in, _) =>
```
s1ck: `expr` or `eExpr` or `e` or `entity`

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
```diff
@@ -180,7 +182,7 @@ trait RelationalCypherRecords[T <: FlatRelationalTable[T]] extends CypherRecords
     from(header, table.distinct)
   }
 
-  def distinct(fields: Var*): R = {
+  def distinct(fields: EntityExpr*): R = {
```
s1ck: I think this is confusing. As mentioned above, I would use `EntityExpr extends Var` instead

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
```diff
@@ -53,7 +52,7 @@ case class RecordHeader(exprToColumn: Map[Expr, String]) {
 
   def expressions: Set[Expr] = exprToColumn.keySet
 
-  def vars: Set[Var] = expressions.collect { case v: Var => v }
+  def vars: Set[Var] = expressions.flatMap(_.owner()).collect { case v: Var => v }
```
s1ck: So, this call potentially returns `Var`s that are not contained in the header but in the expression only?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
```diff
@@ -71,13 +70,19 @@ case class RecordHeader(exprToColumn: Map[Expr, String]) {
     case _ => exprToColumn.getOrElse(expr, throw IllegalArgumentException(s"Header does not contain a column for $expr.\n\t${this.toString}"))
   }
 
-  def ownedBy(expr: Var): Set[Expr] = {
-    exprToColumn.keys.filter(e => e.owner.contains(expr)).toSet
+  def ownedBy(expr: EntityExpr): Set[Expr] = {
+    val members = exprToColumn.keys.filter(e => e.owner.contains(expr)).toSet
+
+    members.flatMap {
```
s1ck: Could you please elaborate on the conceptual idea behind this?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
```diff
@@ -71,13 +70,19 @@ case class RecordHeader(exprToColumn: Map[Expr, String]) {
...
+      case e: EntityExpr if e==expr => Seq(e)
+      case e: EntityExpr => ownedBy(e) + e
+      case other => Seq(other)
+    }
   }
 
   def expressionsFor(expr: Expr): Set[Expr] = {
     expr match {
-      case v: Var => ownedBy(v)
+      case v: EntityExpr => if (exprToColumn.contains(v)) ownedBy(v) + v else ownedBy(v)
```
s1ck: I was thinking that in the `true` case the result is always `Set(v)` according to the new `ownedBy` implementation. This means that for node, rel and path entites, the `expressionsFor` only returns themselves? I'm having problems to understand in which cases the parent entity is in the header and in which cases it is not.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
```diff
@@ -87,14 +92,14 @@ case class RecordHeader(exprToColumn: Map[Expr, String]) {
     exprToColumn.collect { case (k, v) if v == column => k }.toSet
   }
 
-  def aliasesFor(expr: Expr): Set[Var] = {
-    val aliasesFromHeader: Set[Var] = getColumn(expr) match {
+  def aliasesFor(expr: Expr): Set[EntityExpr] = {
```
s1ck: I think this needs to return `Set[Var]` to be semantically correct

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
DarthMax: This function now has to work recursive in order to capture the path case again.
Given a Path `p` with a segment that represents a node, when I ask for all things owned by `p` I also want to get all the things owned by the node contained in the path.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
DarthMax: There is a corner case where this is not true. 
Again given a path `p` and path segment `1` that represents a node. When I now ask for the `expressionsFor` `1`. If internally we would only call `ownedBy`  that would only return the children of one because `1` does not own itself (because it is owned by `p`)

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/table/RecordHeader.scala
DarthMax: I guess that depends on how you want to use that function. A node can have two references on through a `Var` and one through a `PathSegment`. Is the `PathSegment` an alias of the node?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
DarthMax: We also discussed that and it seems to be the better way round.

Mats-SX:
- Rename PathSegment to ListItem
- Drop nodes from the computation
- Return list of rels instead of path

DarthMax:
New Var hierarchy:

```
trait Var
  |__ case class ListSegment
  |__ trait ReturnItem
           |__ NodeVar
           |__ RelationshipVar
           |__ ScalarVar
```

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
-final case class Var(name: String)(val cypherType: CypherType = CTWildcard) extends Expr {
+trait Var extends Expr {
+  def name: String
+}
+
+object Var {
+  def apply(name: String)(cypherType: CypherType = CTWildcard): Var = cypherType.material match {
+    case n: CTNode => NodeVar(name)(n)
+    case r: CTRelationship => RelationshipVar(name)(r)
+    case _ => SimpleVar(name)(cypherType)
```
pstutz: `ListSegment` case not covered

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
+}
+
+trait ReturnItem extends Var
+
+final case class NodeVar(name: String)(val cypherType: CTNode = CTNode) extends ReturnItem {
+
+  override type This = NodeVar
 
-  type This = Var
+  override def owner(): Option[Var] = Some(this)
```
pstutz: Specialise return type

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
+
+  override def withoutType: String = s"$name"
+}
+
+final case class RelationshipVar(name: String)(val cypherType: CTRelationship = CTRelationship) extends ReturnItem {
+
+  override type This = RelationshipVar
 
-  override def withOwner(v: Var): Var = v
+  override def owner(): Option[Var] = Some(this)
```
pstutz: Specialise return type

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
+    case _ => SimpleVar(name)(cypherType)
+  }
+
+  def unapply(arg: Var): Option[String] = Some(arg.name)
+}
+
+final case class ListSegment(index: Int, listVar: Var)(val cypherType: CypherType= CTWildcard) extends Var {
+  override type This = ListSegment
+
+  override def owner(): Option[Var] = Some(listVar)
```
pstutz: Specialise return type

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
+    }
+  }
   override def withoutType: String = s"$name"
+}
+
+final case class SimpleVar(name: String)(val cypherType: CypherType) extends ReturnItem {
 
+  override type This = SimpleVar
+
+  override def owner(): Option[Var] = Some(this)
```
pstutz: Specialise return type

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
pstutz: Remove parentheses

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -97,7 +157,7 @@ final case class StartNode(rel: Expr)(val cypherType: CypherType = CTWildcard) e
 
   override def toString = s"source($rel)"
 
-  override def owner: Option[Var] = rel match {
+  override def owner(): Option[Var] = rel match {
```
pstutz: Remove parentheses for all these calls

okapi-ir/src/test/scala/org/opencypher/okapi/ir/api/expr/ExprTest.scala
```diff
@@ -26,14 +26,14 @@
...
-import org.opencypher.okapi.api.types.{CTNode, CTRelationship, CTWildcard}
+import org.opencypher.okapi.api.types._
 import org.scalatest.{FunSuite, Matchers}
 
 class ExprTest extends FunSuite with Matchers {
 
   test("expressions ignore cypher type in equality") {
-    val n = Var("a")(CTNode)
-    val r = Var("a")(CTRelationship)
+    val n = Var("a")(CTInteger)
```
pstutz: Why was this changed?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/util/VarConverters.scala
```diff
@@ -44,6 +44,10 @@ object VarConverters {
 
   implicit def toVar(s: Symbol): Var = Var(s.name)()
 
+  def toNodeVar(s: Symbol): Var = NodeVar(s.name)()
```
pstutz: Probably not worth it anymore vs `Var("n")(CTNode)`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/util/VarConverters.scala
```diff
@@ -44,6 +44,10 @@ object VarConverters {
...
+
+  def toRelVar(s: Symbol): Var = RelationshipVar(s.name)()
```
pstutz: Probably not worth it anymore vs `Var("r")(CTRelationship)`

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOptimizer.scala
```diff
@@ -62,10 +62,16 @@ object LogicalOptimizer extends DirectCompilationStage[LogicalOperator, LogicalO
   }
 
   def discardScansForNonexistentLabels: PartialFunction[LogicalOperator, LogicalOperator] = {
-    case scan@NodeScan(v, in, _) =>
+    case scan@NodeScan(entityExpr, in, _) =>
       def graphSchema = in.graph.schema
 
-      def emptyRecords = EmptyRecords(Set(v), in, scan.solved)
+      def emptyRecords = {
```
pstutz: The var has to be a `NodeVar` for a node scan, so probably makes sense to tighten the type and remove this match.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
```diff
@@ -91,12 +91,14 @@ trait RelationalCypherRecords[T <: FlatRelationalTable[T]] extends CypherRecords
     val headerWithAliases = header.withAlias(aliasExprs: _*)
 
     val selectHeader = headerWithAliases.select(allExprs: _*)
-    val logicalColumns = allExprs.collect { case e: Var => e.withoutType }
+    val logicalColumns = allExprs.flatMap(_.owner).collect{
```
pstutz: Probably makes sense to add a header operation that collects all return items and use that.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -79,16 +80,75 @@ final case class Param(name: String)(val cypherType: CypherType = CTWildcard) ex
...
+  def apply(name: String)(cypherType: CypherType = CTWildcard): Var = cypherType.material match {
+    case n: CTNode => NodeVar(name)(n)
+    case r: CTRelationship => RelationshipVar(name)(r)
+    case _ => SimpleVar(name)(cypherType)
+  }
+
+  def unapply(arg: Var): Option[String] = Some(arg.name)
+}
+
+final case class ListSegment(index: Int, listVar: Var)(val cypherType: CypherType= CTWildcard) extends Var {
```
pstutz: Maybe rename to `ListElement`

okapi-relational/src/test/scala/org/opencypher/okapi/relational/impl/table/RecordHeaderTest.scala
```diff
@@ -282,6 +307,14 @@ class RecordHeaderTest extends BaseTestSuite {
...
     nHeader.relationshipVars should equalWithTracing(Set.empty)
+    relListHeader.relationshipVars should equalWithTracing(Set.empty)
+  }
+
+  it("returns all rel entities") {
+    rHeader.relationshipEntities should equalWithTracing(Set(r))
+    nHeader.relationshipEntities should equalWithTracing(Set.empty)
+    (nHeader ++ relListHeader).relationshipEntities should equalWithTracing(Set(relListSegment))
+    (rHeader ++ relListHeader).relationshipEntities should equalWithTracing(Set(r, relListSegment))
   }
```
pstutz: Add a test where a node/rel is used in a list and then is removed. Does the list still contain the node? Is the list still working? Does this still work consistently when various aliases are present?

pstutz: Consider adding higher-level list operations to the header and write tests for them.

spark-cypher/src/main/scala/org/opencypher/spark/impl/convert/rowToCypherMap.scala
```diff
@@ -106,4 +109,19 @@ final case class rowToCypherMap(exprToColumn: Seq[(Expr, String)]) extends (Row
       case invalidID => throw UnsupportedOperationException(s"CAPSRelationship ID has to be a Long instead of ${invalidID.getClass}")
     }
   }
+
+  private def collectComplexList(row: Row, expr: Var): CypherList = {
```
pstutz: Consider using a higher-level list operation for this.

pstutz:
I really like the `List` abstraction in the header. I think we should try to interact with such header-lists with an intuitive higher-level API instead of manually modifying and collecting elements. Might it be possible to hide the implementation behind an intuitive Scala List/Vector API?

I'm not certain what happens when an item that is used in a list is removed from the header. Given the complexity I think the header should contain a short documentation that explains the behaviour and some tests that ensure it works as described.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
DarthMax: In which case would it create a ListSegment?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
DarthMax: The return type can only be so far specialized as to make it a `ReturnItem`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/util/VarConverters.scala
DarthMax: Prior to this change the `toVar` method was used in many paces (without a CypherType). I wanted to keep this convenience

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOptimizer.scala
DarthMax: Let's tackle these changes in separate PRs whenever we encounter a possibility to restrict types more. These changes often entail more changes in other places and I want to get this PR in soon

okapi-relational/src/test/scala/org/opencypher/okapi/relational/impl/table/RecordHeaderTest.scala
DarthMax: I can currently not think of a situation where a list element ca be removed deliberately without constructing a new list.

PR #520, opened 2018-06-19
Evaluate UnionAll and CartesianProduct via RelationalCypherRecords

s1ck:
* the union now checks for the correct column order of the input records and reorders the right side only if necessary

### DIFF
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
@@ -37,7 +37,7 @@ import org.opencypher.okapi.ir.api.block.{Asc, Desc, SortItem}
 import org.opencypher.okapi.ir.api.expr._
 import org.opencypher.okapi.ir.api.{Label, PropertyKey, RelType}
 import org.opencypher.okapi.relational.api.io.RelationalEntityMapping._
-import org.opencypher.okapi.relational.impl.physical.{Ascending, Descending, JoinType, Order}
+import org.opencypher.okapi.relational.impl.physical._
 import org.opencypher.okapi.relational.impl.table.RecordHeader
 
 trait FlatRelationalTable[T <: FlatRelationalTable[T]] extends CypherTable {
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
@@ -172,7 +172,12 @@ trait RelationalCypherRecords[T <: FlatRelationalTable[T]] extends CypherRecords
   }
 
   def unionAll(other: R): R = {
-    val unionData = table.unionAll(other.table)
+    val orderedTable = if (table.physicalColumns != other.table.physicalColumns) {
+      other.table.select(table.physicalColumns: _*)
+    } else {
+      other.table
+    }
+    val unionData = table.unionAll(orderedTable)
     from(header, unionData)
   }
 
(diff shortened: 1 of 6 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/CAPSTable.scala
```diff
@@ -105,20 +105,28 @@ object SparkCypherTable {
...
+          val joinExpr = joinCols.map {
+            case (l, r) => df.col(l) === other.df.col(r)
+          }.reduce((acc, expr) => acc && expr)
+
+          df.join(other.df, joinExpr, joinTypeString)
+//          // TODO: the join produced corrupt data when the previous operator was a cross. We work around that by using a
+//          // subsequent select. This can be removed, once https://issues.apache.org/jira/browse/SPARK-23855 is solved or we
+//          // upgrade to Spark 2.3.0
+//          val potentiallyCorruptedResult = df.join(other.df, joinExpr, joinTypeString)
+//          potentiallyCorruptedResult.select("*")
```
Mats-SX: I guess we can remove this comment now then?

spark-cypher/src/main/scala/org/opencypher/spark/api/io/CAPSTable.scala
```diff
@@ -105,20 +105,28 @@ object SparkCypherTable {
         case LeftOuterJoin => "left_outer"
         case RightOuterJoin => "right_outer"
         case FullOuterJoin => "full_outer"
+        case CrossJoin => "cross"
```
Mats-SX: Perhaps we should move this conversion to `JoinType` itself, like `.name` or something.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
```diff
@@ -172,7 +172,12 @@ trait RelationalCypherRecords[T <: FlatRelationalTable[T]] extends CypherRecords
   }
 
   def unionAll(other: R): R = {
-    val unionData = table.unionAll(other.table)
+    val orderedTable = if (table.physicalColumns != other.table.physicalColumns) {
```
Mats-SX: Should we do a check here that if the physical columns are more different than just ordered differently, we will fail before Spark does? We could also rely on that invariant being checked elsewhere, but I'm not sure if it is.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/CAPSTable.scala
s1ck: Yep, just running a CI build to see if it's not needed anymore.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
s1ck: It's not checked elsewhere and Catalyst would fail in the first planning phase. But we can also check it here as not all backends might act in such a sensible way.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/CAPSTable.scala
s1ck: Turns out it is still required: https://build.neohq.net/viewLog.html?buildId=3768261&buildTypeId=OpenCypher_CasBuildPr

spark-cypher/src/main/scala/org/opencypher/spark/api/io/CAPSTable.scala
Mats-SX: Are they really spark-specific? They seems to be common sense names to me. But we could do it with an implicit class and call it `sparkName` as well... doesn't matter a lot.

PR #525, opened 2018-06-21
Scaling neo4j pgds

Mats-SX:
Also contains #527


Mats-SX:
@DarthMax Because we do not yet have a fallback for the case when the schema procedure is not available: https://github.com/opencypher/cypher-for-apache-spark/pull/525/files#diff-84eb98f4d849f56b66e9c428762029ddR93

I think we can reuse the schema inference code we already have to do the fallback, with some modifications.

Another reason is because removing `CommunityNeo4jPGDS` would be a surface change, and this new PGDS is not interfaced the same way (with Cypher queries).

PR #526, opened 2018-06-21
Use RelationalCypherRecords throughout physical planning

s1ck:


spark-cypher/src/main/scala/org/opencypher/spark/impl/physical/operators/CAPSPhysicalOperator.scala
```diff
@@ -26,75 +26,31 @@
...
 
   override def args: Iterator[Any] = super.args.flatMap {
     case RecordHeader | Some(RecordHeader) => None
-    case other                               => Some(other)
+    case other => Some(other)
   }
 }
-
-object CAPSPhysicalOperator {
-  def joinRecords(
```
Mats-SX: So this functionality now lives on `DataFrameTable` instead?

PR #527, opened 2018-06-21
Fix bug where schema concatenation did not resolve relationship properties correctly

Mats-SX:


okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
```diff
@@ -221,6 +220,15 @@ final case class SchemaImpl(
     }
     val newNodeKeyMap = labelPropertyMap ++ other.labelPropertyMap ++ LabelPropertyMap(nulledOut)
 
+
```
DarthMax: I think it is possible to abstract that resolver function. It seems to be the same one as the one for labels

Mats-SX: Funny you should say that (I anticipated it). It was abstracted some time ago, but that abstraction was so general it was very difficult to understand it. We can do it again and do it better I suppose, but it will likely result in a very generic signature.

s1ck:
Merged with https://github.com/opencypher/cypher-for-apache-spark/pull/525

PR #548, opened 2018-07-09
Merge CAPS physical and OKAPI flat operators into OKAPI relational operators

s1ck:


pstutz:
Need to rebase and investigate the possible Spark bug related to `DF.persist`.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -480,7 +484,7 @@ object SchemaTyper {
         case ToBoolean =>
```
s1ck: maybe we should handle this similar to `toString` and derive the nullability from the inner expression

pstutz: I think it can be null even when a string input was non-null: a string can fail to parse.

s1ck: This worked with the previous version (i.e. similar to `toString`), too:
```
it("toBoolean() on invalid strings") {
      val given = initGraph("CREATE ({id: 'tr ue'}), ({id: 'fa lse'})")

      val result = given.cypher("MATCH (n) RETURN toBoolean(n.id) AS nId")

      result.records.toMaps should equal(
        Bag(
          CypherMap("nId" -> null),
          CypherMap("nId" -> null)
        )
      )
    }
```

pstutz: In that case the result might be correct, but the nullability inferred by the typer would be wrong: it would infer non-nullable because of the non-nullable string arg, but the actual result is null.

PR #555, opened 2018-07-13
Added toBoolean function

grewalr:
Current issue invoking toBoolean with invalid types. Added the following to the blacklisted scenarios:

Feature "TypeConversionFunctions": Scenario "`toBoolean()` on invalid types #1"
Feature "TypeConversionFunctions": Scenario "`toBoolean()` on invalid types #2"
Feature "TypeConversionFunctions": Scenario "`toBoolean()` on invalid types #3"
Feature "TypeConversionFunctions": Scenario "`toBoolean()` on invalid types #4"


okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -245,6 +246,18 @@ object SchemaTyper {
           error(WrongNumberOfArguments(expr, 1, seq.size))
       }
 
+    // TODO: Remove as soon as FunctionInvocationTyper is capable of supporting all Neo4j front end types
+    case expr: FunctionInvocation if expr.function == ToBoolean =>
+      expr.arguments match {
+        case Seq(first) =>
+          for {
+            _ <- process[R](first)
+            existsType <- recordAndUpdate(expr -> CTBoolean)
```
s1ck: I ran into an issue in our refactoring PR for `toString`. This should use the inner types nullability and then use `CTBoolean.asNullableAs(innerType)`

DarthMax: For ToBoolean you can actually skip this and the following code here: https://github.com/opencypher/cypher-for-apache-spark/blob/9941ff0e890b8dc8ce06f76e171aa8753375d3bd/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala#L474-L494

```scala

        case ToBoolean =>
          pure[R, Set[(Seq[CypherType], CypherType)]](
            Set(
              Seq(CTString) -> CTBoolean,
              Seq(CTBoolean) -> CTBoolean
            ))
```

s1ck: You're probably right. For `toString` that didn't work since type signatures for `ToString` contained the temporal types which we can't handle atm.

s1ck: Maybe this also fixes the now blacklisted scenario.

DarthMax: I tried that but we need support for list index access for these TCK tests. i"m implementing that atm

PR #568, opened 2018-07-26
Implement list index access

DarthMax:
Implement

```
WITH [1,2,3] as list
UNWIND [0,1,2] as i
RETURN list[i] as item
```


spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -218,6 +218,15 @@ object SparkSQLExprMapper {
           }
           caseColumn
 
+        case ContainerIndex(container, index) =>
+          val indexCol = index.asSparkSQLExpr
+          val containerCol = container.asSparkSQLExpr
+
+          container.cypherType.material match {
+            case _: CTList => containerCol.get(indexCol)
+            case other => throw NotImplementedException(s"Support for accessing $other by index is not supported")
```
pstutz: "Accessing $other by index is not supported"

DarthMax: :+1:

PR #570, opened 2018-07-26
Added toInteger fn

grewalr:


Mats-SX:
@grewalr Why is there three commits with the message `Added toInteger fn` and two commits with the message `Removed passed scenarios from blacklist`? It seems like a consequence of some merging/rebasing; can this be cleaned up?

grewalr:
Hmm seems like a rebase/merge commit has gone wrong somewhere. I'll take a look at aggregating those commits.

PR #571, opened 2018-07-27
Implemented toInteger function

grewalr:


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/FunctionsBehaviour.scala
```diff
@@ -449,4 +449,59 @@ class FunctionsBehaviour extends CAPSTestSuite with DefaultGraphInit {
...
+
+      result.getRecords.toMaps should equal(
+        Bag(
+          CypherMap("age" -> 42)
+        )
+      )
+    }
+
+    it("toInteger() on float") {
+      val given = initGraph("CREATE (:Person {weight: '82.9'})")
```
DarthMax: This will give you a string "82.9" not a Float, won't it?

PR #576, opened 2018-07-31
Add Neo4jViewPGDS

pstutz:
Co-Authored-By: Martin Junghanns <s1ck@users.noreply.github.com>

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2016-2018 "Neo4j Sweden, AB" [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Attribution Notice under the terms of the Apache License 2.0
+ *
+ * This work was created by the collective efforts of the openCypher community.
+ * Without limiting the terms of Section 6, any Derivative Work that is not
+ * approved by the public consensus process of the openCypher Implementers Group
+ * should not be described as “Cypher” (and Cypher® is a registered trademark of
+ * Neo4j Inc.) or as "openCypher". Extensions by implementers or prototypes or
+ * proposals for change that have been documented or implemented should only be
+ * described as "implementation extensions to Cypher" or as "proposed changes to
+ * Cypher that are not yet approved by the openCypher community".
+ */
+package org.opencypher.spark.api.io.neo4j
+
+import org.opencypher.okapi.api.graph.{GraphName, PropertyGraph}
+import org.opencypher.okapi.api.io.PropertyGraphDataSource
+import org.opencypher.okapi.api.schema.Schema
+import org.opencypher.okapi.impl.exception.UnsupportedOperationException
+import org.opencypher.spark.api.CAPSSession
+import org.opencypher.spark.api.io.neo4j.Neo4jPropertyGraphDataSource.defaultEntireGraphName
+
+case class Neo4jViewPropertyGraphDataSource(
+  viewQuery: String,
+  config: Neo4jConfig,
+  maybeSchema: Option[Schema] = None,
+  omitImportFailures: Boolean = false,
+  entireGraphName: GraphName = defaultEntireGraphName
+)(implicit session: CAPSSession) extends PropertyGraphDataSource {
+
+  private val wrappedNeo4jDs = Neo4jPropertyGraphDataSource(
+    config, omitImportFailures = omitImportFailures, entireGraphName = entireGraphName)
+
+  override def hasGraph(name: GraphName): Boolean = wrappedNeo4jDs.hasGraph(name)
+
+  override def graph(name: GraphName): PropertyGraph = {
+    wrappedNeo4jDs.graph(name).cypher(viewQuery).graph
+  }
+
+  override def schema(name: GraphName): Option[Schema] = maybeSchema
+
+  override def store(name: GraphName, graph: PropertyGraph): Unit =
+    throw UnsupportedOperationException("Storing a view is not supported")
+
+  override def delete(name: GraphName): Unit = throw UnsupportedOperationException("Deleting a view is not supported")
+
+  override val graphNames: Set[GraphName] = wrappedNeo4jDs.graphNames
+
+}
--- a/okapi-testing/src/main/scala/org/opencypher/okapi/testing/PGDSAcceptance.scala
+++ b/okapi-testing/src/main/scala/org/opencypher/okapi/testing/PGDSAcceptance.scala
@@ -35,6 +35,7 @@ import org.opencypher.okapi.impl.exception.{GraphAlreadyExistsException, GraphNo
 import org.opencypher.okapi.testing.Bag._
 import org.opencypher.okapi.testing.propertygraph.{CreateGraphFactory, InMemoryTestGraph}
 import org.scalatest.BeforeAndAfterEach
+import org.opencypher.okapi.impl.exception.UnsupportedOperationException
 
 import scala.util.{Failure, Success, Try}
 
@@ -312,23 +313,4 @@ trait PGDSAcceptance[Session <: CypherSession] extends BeforeAndAfterEach {
     }
   }
 
-}
-
-case class SingleGraphDataSource(graphName: GraphName, graph: PropertyGraph) extends PropertyGraphDataSource {
-
-  override def hasGraph(name: GraphName): Boolean = {
-    name == graphName
-  }
-
-  override def graph(name: GraphName): PropertyGraph = {
-    if (name == graphName) graph else throw GraphNotFoundException(s"Graph $name not found")
-  }
-
-  override def schema(name: GraphName): Option[Schema] = ???
-
-  override def store(name: GraphName, graph: PropertyGraph): Unit = ???
-
-  override def delete(name: GraphName): Unit = ???
-
-  override def graphNames: Set[GraphName] = ???
 }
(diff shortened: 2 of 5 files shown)


s1ck:
@alastai This is a first approach for a view based Neo4j PGDS using Cypher 10 syntax to create the view. The following example demonstrates the usage. We can use this for further discussions.

```scala
  session.registerSource(Namespace("neo"), GraphSources.cypher.neo4j(neo4j.dataSourceConfig))
  session.registerSource(Namespace("fs"), GraphSources.fs(rootPath = getClass.getResource("/csv").getFile).csv)

  // Copy the products from filesystem (csv) to Neo4j
  session.cypher(
    """
      |CREATE GRAPH neo.products {
      | FROM GRAPH fs.products
      | RETURN GRAPH
      |}
    """.stripMargin)

  // Create a view on top of the Neo4j graph that just contains DVDs and customers who bought them
  val viewPGDS = GraphSources.cypher.neo4jView(
    """
      |MATCH (p:Product)
      |WHERE p.category = 'DVD'
      |OPTIONAL MATCH (p)<-[b:BOUGHT]-(c:Customer)
      |CONSTRUCT
      |  CLONE p, b, c
      |  NEW (p)<-[b]-(c)
      |RETURN GRAPH
    """.stripMargin, neo4j.dataSourceConfig)

  session.registerSource(Namespace("DVD_view"), viewPGDS)

  // Return all nodes within the DVD view
  session.cypher(
    """
      |FROM GRAPH DVD_view.products
      |MATCH (n) RETURN n
      |ORDER BY labels(n), n.name, n.title
    """.stripMargin).show

```

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import org.opencypher.okapi.api.graph.{GraphName, PropertyGraph}
+import org.opencypher.okapi.api.io.PropertyGraphDataSource
+import org.opencypher.okapi.api.schema.Schema
+import org.opencypher.okapi.impl.exception.UnsupportedOperationException
+import org.opencypher.spark.api.CAPSSession
+import org.opencypher.spark.api.io.neo4j.Neo4jPropertyGraphDataSource.defaultEntireGraphName
+
+case class Neo4jViewPropertyGraphDataSource(
+  viewQuery: String,
```
Mats-SX: So this PGDS stores a single view which is evaluated over a base graph determined by the `FROM` clause in a query?

```
FROM foo.bar
MATCH ...
// etc
```

if `foo` is registered for a `Neo4jViewPGDS` it will then use `bar` as the graph to evaluate the pre-registered query over.

This design indicates that if many views are interesting, many PGDSs will have to be registered. This is not a flaw, but I can see an alternative design where view queries are managed in some `Map[GraphName, String]` field. The difference in that design is that since the graph name is used for view selection, it couldn't be used for base graph selection => the views would need to always be evaluated over the entire Neo4j graph.

One side effect of the implemented design that we might want to think about is this:

If I (like in the example above) register the same Neo4j instance to both a 'standard' Neo4j PGDS (at `neo4j`) and a ViewPGDS (at `view1`) and then store a graph `neo4j.myGraph`, then I will be able to refer to `view1.myGraph`, but this will not be the same graph; this is `neo4j.myGraph` plus the view query defined for `view1`.

This isn't a huge problem since the namespaced names are different (meaning they are different graphs even though they share a part of the name), but we need to make users understand this.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
```diff
@@ -0,0 +1,62 @@
...
+  override def hasGraph(name: GraphName): Boolean = wrappedNeo4jDs.hasGraph(name)
+
+  override def graph(name: GraphName): PropertyGraph = {
+    wrappedNeo4jDs.graph(name).cypher(viewQuery).graph
+  }
+
+  override def schema(name: GraphName): Option[Schema] = maybeSchema
+
+  override def store(name: GraphName, graph: PropertyGraph): Unit =
+    throw UnsupportedOperationException("Storing a view is not supported")
```
Mats-SX: I guess we could support this operation. The semantics are a bit strange, but they aren't completely unreasonable. If we supported this as just delegating to the `wrappedNeo4jDs` it would mean that the provided `graph` becomes available as a base graph under the same `QualifiedGraphName`.

`CREATE GRAPH view1.foo { ... }` <-- stores some graph in the underlying Neo4j db
`FROM GRAPH view1.foo MATCH ...` <-- reads that graph and evaluates the view query over it before matching

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def delete(name: GraphName): Unit = throw UnsupportedOperationException("Deleting a view is not supported")
```
Mats-SX: Similarly as for `.store` we could allow this. OTOH perhaps there's a value in making this PGDS purely read-only.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
pstutz: I added more documentation to explain the usage of the DS.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
pstutz: Given the design we discussed this morning I think it's better to not support these operation via a view, but for the user to use the underlying DS for those operations.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/Neo4jViewPropertyGraphDataSource.scala
pstutz: Yes, I like read-only for now.

pstutz:
The new design allows to construct a view on top of any graph in the Catalog.

Usage:
```scala
  // Create a view that contains only DVDs and customers who bought them
  val viewPGDS = GraphSources.cypher.view(
    """
      |MATCH (p:Product)
      |WHERE p.category = 'DVD'
      |OPTIONAL MATCH (p)<-[b:BOUGHT]-(c:Customer)
      |CONSTRUCT
      |  NEW (p)<-[b]-(c)
      |RETURN GRAPH
    """.stripMargin)

  session.registerSource(Namespace("DVD_view"), viewPGDS)

    // Return all nodes in the `neo.products` graph when viewed via the DVD view
  session.cypher(
    """
      |FROM GRAPH DVD_view.neo.products
      |MATCH (n) RETURN n
      |ORDER BY labels(n), n.name, n.title
    """.stripMargin).show
```

API:
```scala
  /**
    * Creates a data source that can construct a view defined by a query over any other graph stored in the catalog.
    *
    * Example:
    *  - given a graph with qualified graph name `foo.bar`
    *  - given a view with namespace `view` and no custom mapping for `foo.bar`
    *
    * In this case the qualified graph name of the view of that graph is `view.foo.bar`.
    *
    * A view query is a query that uses Cypher 10 to construct a graph. This is an example for a view query that returns
    * the entire underlying graph without any changes:
    * {{{
    *   MATCH (n)
    *   MATCH (s)-[r]->(t)
    *   CONSTRUCT
    *     CLONE n
    *     NEW (s)-[r]->(t)
    *   RETURN GRAPH
    * }}}
    *
    * @param viewQuery               query that constructs the view
    * @param customGraphNameMappings allows to customise the mapping of graph names to QGNs of underlying graphs
    * @return view property graph data source
    */
  def view(viewQuery: String, customGraphNameMappings: Map[GraphName, QualifiedGraphName] = Map.empty)
    (implicit session: CAPSSession): ViewPropertyGraphDataSource = {
    ViewPropertyGraphDataSource(viewQuery)
  }
```

pstutz:
We might want to move the view concept/PGDS to Okapi API and add syntactic sugar for their creation to the session.

pstutz:
Usage with syntactic sugar for creating & registering:
```scala
  // Create a view that contains only DVDs and customers who bought them
   session.registerView(Namespace("DVD_view"),
     """
       |MATCH (p:Product)
       |WHERE p.category = 'DVD'
       |OPTIONAL MATCH (p)<-[b:BOUGHT]-(c:Customer)
       |CONSTRUCT
       |  NEW (p)<-[b]-(c)
       |RETURN GRAPH
     """.stripMargin)

   // Return all nodes in the `neo.products` graph when viewed via the DVD view
   session.cypher(
     """
       |FROM GRAPH DVD_view.neo.products
       |MATCH (n) RETURN n
       |ORDER BY labels(n), n.name, n.title
     """.stripMargin).show
```

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/view/ViewPGDSAcceptanceTest.scala
```diff
@@ -0,0 +1,66 @@
...
+    val graph: CAPSScanGraph = CAPSScanGraphFactory(testGraph)
+    sessionDs.store(GraphName("test"), graph)
+    val viewDs = GraphSources.view(
+      """
+        |MATCH (n)
+        |MATCH (s)-[r]->(t)
+        |CONSTRUCT
+        |  CLONE n
+        |  NEW (s)-[r]->(t)
+        |RETURN GRAPH
```
Mats-SX: This is essentially a pass-through view, isn't it? Like, include everything from the underlying graph.

pstutz: Yes, it just runs the acceptance tests via a pass-though view.

okapi-api/src/main/scala/org/opencypher/okapi/api/io/ViewPropertyGraphDataSource.scala
```diff
@@ -0,0 +1,87 @@
...
+  *
+  * It interprets the graph name passed to it as the qualified graph name of the underlying graph to which the view is
+  * applied.
+  *
+  * @param viewQuery query applied to the underlying graph
+  * @param customGraphNameMappings allows to customise the mapping of view names to QGNs
+  */
+case class ViewPropertyGraphDataSource(
+  viewQuery: String,
+  customGraphNameMappings: Map[GraphName, QualifiedGraphName] = Map.empty
```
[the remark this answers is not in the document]
pstutz: Yes, it was required to get the acceptance test with pass-though view working.

okapi-api/src/main/scala/org/opencypher/okapi/api/graph/CypherSession.scala
```diff
@@ -70,6 +70,33 @@ trait CypherSession {
...
+    *   CONSTRUCT
+    *     CLONE n
+    *     NEW (s)-[r]->(t)
+    *   RETURN GRAPH
+    * }}}
+    *
+    * @param namespace namespace where the view data source is registered
+    * @param viewQuery query that constructs the view
+    */
+  def registerView(namespace: Namespace, viewQuery: String): Unit = {
```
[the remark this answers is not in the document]
pstutz: Yes, putting it at the end might be nicer. Doing it like this requires less magic though, so I don't see changing it as a clear-cut win. I suggest we await Alastair's feedback on this.

PR #581, opened 2018-08-03
Stop planning unnecessary property projections

### LINKED ISSUE #580: Avoid projecting property expressions where possible

s1ck:
* fixes #580

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/PredicateBehaviour.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/PredicateBehaviour.scala
@@ -35,6 +35,14 @@ import org.scalatest.DoNotDiscover
 @DoNotDiscover
 class PredicateBehaviour extends CAPSTestSuite with DefaultGraphInit {
 
+  it("can evaluate predicates on non-existing properties") {
+    val given = initGraph("CREATE ()")
+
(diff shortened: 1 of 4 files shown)


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/PredicateBehaviour.scala
```diff
@@ -35,6 +35,14 @@ import org.scalatest.DoNotDiscover
 @DoNotDiscover
 class PredicateBehaviour extends CAPSTestSuite with DefaultGraphInit {
 
+  it("can evaluate predicates on non-existing properties") {
```
Mats-SX: Was this test case red before the fix?

s1ck: No, it's an additional test that verifies we correctly evaluate predicate expressions based on non-existing properties.

s1ck: Before we have just one spark-cypher-test (`can evaluate coalesce on non-existing expressions`) that implicitly tests for that.

Mats-SX: Very nice.

PR #582, opened 2018-08-03
Add rewriter that replaces cross joins with value joins

s1ck:
* needs further changes to tree framework (i.e. stateful rewriting)


s1ck:
Requires merging https://github.com/opencypher/cypher-for-apache-spark/pull/583

PR #598, opened 2018-08-08
Turn node/rel property key maps into simple Scala maps

pstutz:
Provide extra functionality with extension methods

Also add a test for Schema JSON serialisation/deserialisation


s1ck:
cya `map.map.map` :wave:

PR #600, opened 2018-08-09
Update frontend version

pstutz:
Co-Authored-By: Mats Rydberg <mats@neotechnology.com>

### DIFF
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -29,8 +29,8 @@ package org.opencypher.okapi.ir.impl
 import cats.implicits._
 import org.atnos.eff._
 import org.atnos.eff.all._
-import org.opencypher.v9_1.ast
-import org.opencypher.v9_1.{expressions => exp}
+import org.opencypher.v9_0.ast
+import org.opencypher.v9_0.{expressions => exp}
 import org.opencypher.okapi.api.graph.QualifiedGraphName
 import org.opencypher.okapi.api.schema.Schema
 import org.opencypher.okapi.api.types._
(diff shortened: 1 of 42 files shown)


okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -159,17 +159,18 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement[Expr],
           }
         } yield refs
 
-      case ast.With(distinct, ast.ReturnItems(_, items), _, _, _, None) =>
+      case ast.With(distinct, ast.ReturnItems(_, items), _, _, _, where) =>
         for {
           fieldExprs <- items.toList.traverse(convertReturnItem[R])
+          given <- convertWhere(where)
```
Mats-SX: Wow, this looks like a nice feature win. I'm somewhat surprised we didn't see any TCK scenarios be enabled by this.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -82,8 +82,8 @@ class SchemaTyperTest extends BaseTestSuite with Neo4jAstTestSupport with Mockit
     implicit val context = typeTracker("a" -> CTNode)
 
     assertExpr.from("count(*)") shouldHaveInferredType CTInteger
-    assertExpr.from("count(a)") shouldHaveInferredType CTInteger
-    assertExpr.from("count(a.name)") shouldHaveInferredType CTInteger
+    assertExpr.from("count(a)") shouldHaveInferredType CTInteger.nullable
+    assertExpr.from("count(a.name)") shouldHaveInferredType CTInteger.nullable
```
Mats-SX: I guess these are regressions stemming from us always considering the output of a function nullable. These functions will never yield null.

PR #601, opened 2018-08-09
Expose scans

DarthMax:
This removes the relational scan operator and thus exposes scan operations in the query plan


spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/support/creation/caps/SingleTableGraphFactory.scala
```diff
@@ -48,9 +48,7 @@ object SingleTableGraphFactory extends CAPSTestGraphFactory {
     val joinExpr: (Expr, Expr) = Var("n")(CTNode) -> relsStartOp.header.startNodeFor(Var("r")(CTRelationship))
     val joinOp = Join(nodesStartOp, relsStartOp, Seq(joinExpr), LeftOuterJoin)
 
-    val baseTable = caps.records.from(joinOp.header, joinOp.table)
-
-    caps.graphs.singleTableGraph(baseTable, scanGraph.schema, Set(0))(caps.basicRuntimeContext())
+    caps.graphs.singleTableGraph(joinOp, scanGraph.schema, Set(0))(caps.basicRuntimeContext())
```
pstutz: It should probably be possible to pass in the used tags.

PR #608, opened 2018-08-13
Fix bug where numbers would get automatically coerced to strings

Mats-SX:
The bug is due to UNION semantics in Spark, but we avoid this operation
by manually checking the DataFrame schemas before UNION.


spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/support/creation/caps/SingleTableGraphFactory.scala
```diff
@@ -39,14 +39,11 @@ object SingleTableGraphFactory extends CAPSTestGraphFactory {
 
   override def apply(propertyGraph: InMemoryTestGraph)(implicit caps: CAPSSession): RelationalCypherGraph[DataFrameTable] = {
     val scanGraph = CAPSScanGraphFactory(propertyGraph)
-    val nodes = scanGraph.nodes("n")
-    val rels = scanGraph.relationships("r")
+    val nodesOp = scanGraph.scanOperator(CTNode)
+    val relsOp = scanGraph.scanOperator(CTRelationship)
```
DarthMax: :+1:

PR #615, opened 2018-08-15
Add mathematical (logarithmic) functions

s1ck:


spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -44,6 +44,8 @@ object SparkSQLExprMapper {
 
   private val FALSE_LIT: Column = functions.lit(false)
 
+  private val E: Column = functions.lit(2.718281828459045)
```
DarthMax: :stuck_out_tongue_winking_eye:  Nice

s1ck: c&p neo output :see_no_evil:

PR #617, opened 2018-08-16
Tree improvements

pstutz:
Supports improved children detection:
```scala
  abstract class ComplexExample extends AbstractTreeNode[ComplexExample]

  case class SimpleA() extends ComplexExample

  case class SimpleB() extends ComplexExample

  case class SimpleC() extends ComplexExample

  case class SimpleD() extends ComplexExample

  case class Multi(
    ints: NonEmptyList[Integer],
    as: NonEmptyList[SimpleA],
    s: List[String],
    bs: List[SimpleB],
    maybeL: Option[Long],
    maybeC: Option[SimpleC],
    maybeD: Option[SimpleD]
  ) extends ComplexExample

it("can infer children for complex case classes") {
    val instance = Multi(
      NonEmptyList.one(1),
      NonEmptyList.of(SimpleA(), SimpleA()),
      List("A", "B"),
      List(SimpleB(), SimpleB()),
      Some(1L),
      None,
      Some(SimpleD())
    )

    instance.children.toList should equal(List(
      SimpleA(), SimpleA(), SimpleB(), SimpleB(), SimpleD()
    ))

    instance.withNewChildren(Array(
      SimpleA(), SimpleC()
    )) should equal(Multi(
      NonEmptyList.one(1),
      NonEmptyList.of(SimpleA()),
      List("A", "B"),
      List(),
      Some(1L),
      Some(SimpleC()),
      None
    ))
  }
```

### DIFF
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/CypherStatement.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/CypherStatement.scala
@@ -27,45 +27,46 @@
 package org.opencypher.okapi.ir.api
 
 import org.opencypher.okapi.ir.api.block.{Binds, Block}
+import org.opencypher.okapi.ir.api.expr.Expr
 
-sealed trait CypherStatement[E] extends Block[E] {
+sealed trait CypherStatement extends Block {
   def info: QueryInfo
 }
 
-final case class CypherQuery[E](
+final case class CypherQuery(
     info: QueryInfo,
-    model: QueryModel[E]
-) extends CypherStatement[E] {
-  override def after: List[Block[E]] = model.after
+    model: QueryModel
+) extends CypherStatement {
+  override def after: List[Block] = model.after
 
-  override def binds: Binds[E] = model.binds
+  override def binds: Binds = model.binds
 
-  override def where: Set[E] = model.where
+  override def where: Set[Expr] = model.where
 
   override def graph: IRGraph = model.graph
 }
 
-final case class CreateGraphStatement[E](
+final case class CreateGraphStatement(
     info: QueryInfo,
     graph: IRGraph,
-    innerQuery: CypherQuery[E]
-) extends CypherStatement[E] {
+    innerQuery: CypherQuery
+) extends CypherStatement {
 
-  override val after: List[Block[E]] = List(innerQuery)
+  override val after: List[Block] = List(innerQuery)
 
-  override def binds: Binds[E] = Binds.empty
+  override def binds: Binds = Binds.empty
 
-  override def where: Set[E] = Set.empty
+  override def where: Set[Expr] = Set.empty
 }
 
-final case class DeleteGraphStatement[E](
+final case class DeleteGraphStatement(
   info: QueryInfo,
   graph: IRGraph
-) extends CypherStatement[E] {
+) extends CypherStatement {
 
-  override val after: List[Block[E]] = List.empty
+  override val after: List[Block] = List.empty
 
-  override def binds: Binds[E] = Binds.empty
+  override def binds: Binds = Binds.empty
 
-  override def where: Set[E] = Set.empty
+  override def where: Set[Expr] = Set.empty
 }
--- a/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala
+++ b/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala
@@ -42,10 +42,12 @@ import org.opencypher.okapi.relational.impl.planning.ConstructGraphPlanner._
 import org.opencypher.okapi.relational.impl.table.RecordHeader
 import org.opencypher.okapi.relational.impl.{operators => relational}
 
+import scala.reflect.runtime.universe.TypeTag
+
 object RelationalPlanner {
 
   // TODO: rename to 'plan'
-  def process[T <: Table[T]](input: LogicalOperator)
+  def process[T <: Table[T] : TypeTag](input: LogicalOperator)
     (implicit context: RelationalRuntimeContext[T]): RelationalOperator[T] = {
 
     implicit val caps: CypherSession = context.session
@@ -191,7 +193,7 @@ object RelationalPlanner {
         val rightWithAliases = rightResult.alias(renameExprs)
         // 2. Drop Join expressions and their children in rhs
         val exprsToRemove = joinExprs.flatMap(v => rightHeader.ownedBy(v))
-        val reducedRhsData = rightWithAliases.drop(exprsToRemove)
+        val reducedRhsData = rightWithAliases.dropExprSet(exprsToRemove)
         // 3. Compute distinct rows in rhs
         val distinctRhsData = relational.Distinct(reducedRhsData, renameExprs.map(_.alias))
         // 4. Join lhs and prepared rhs using a left outer join
@@ -200,7 +202,7 @@ object RelationalPlanner {
         val targetExpr = renameExprs.head.alias
         joinedData.addInto(IsNotNull(targetExpr)(CTBoolean) -> predicateField.targetField)
 
-      case logical.OrderBy(sortItems: Seq[SortItem[Expr]], in, _) =>
+      case logical.OrderBy(sortItems: Seq[SortItem], in, _) =>
         relational.OrderBy(process[T](in), sortItems)
 
       case logical.Skip(expr, in, _) =>
@@ -215,7 +217,7 @@ object RelationalPlanner {
     }
   }
 
-  def planScan[T <: Table[T]](
+  def planScan[T <: Table[T] : TypeTag](
     maybeInOp: Option[RelationalOperator[T]],
     logicalGraph: LogicalGraph,
     entityVar: Var
@@ -236,7 +238,7 @@ object RelationalPlanner {
   }
 
   // TODO: process operator outside of def
-  private def planOptional[T <: Table[T]](lhs: LogicalOperator, rhs: LogicalOperator)
+  private def planOptional[T <: Table[T] : TypeTag](lhs: LogicalOperator, rhs: LogicalOperator)
     (implicit context: RelationalRuntimeContext[T]): RelationalOperator[T] = {
     val lhsOp = process[T](lhs)
     val rhsOp = process[T](rhs)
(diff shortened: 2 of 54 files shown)


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/CypherStatement.scala
```diff
@@ -27,45 +27,46 @@
 package org.opencypher.okapi.ir.api
 
 import org.opencypher.okapi.ir.api.block.{Binds, Block}
+import org.opencypher.okapi.ir.api.expr.Expr
 
-sealed trait CypherStatement[E] extends Block[E] {
+sealed trait CypherStatement extends Block {
```
s1ck: I think at some point there was the idea that IR should also work with other implementations than our `Expr` This is why it's not really "needed".

okapi-trees/src/main/scala/org/opencypher/okapi/trees/AbstractTreeNode.scala
```diff
@@ -242,8 +192,8 @@ object AbstractTreeNode {
       val copyMethodSymbol = tpe.decl(TermName("copy")).asMethod
       instanceMirror.reflectMethod(copyMethodSymbol)
     } catch {
-      case e: Exception =>
-        throw new UnsupportedOperationException(s"Could not reflect the copy method of $instance", e)
+      case e: Exception => throw new UnsupportedOperationException(
+        s"Could not reflect the copy method of ${instance.toString.filterNot(_ == '$')}", e)
```
Mats-SX: What does the `filterNot` ensure here? Are there classes that have a single dollar sign as `toString()`?

okapi-trees/src/main/scala/org/opencypher/okapi/trees/TreeNode.scala
```diff
@@ -172,42 +179,15 @@ abstract class TreeNode[T <: TreeNode[T] : ClassTag] extends Product with Traver
       .filterNot(_.isMethod)
       .toList
       .map(currentMirror.reflect(this).reflectField)
-      .map(termSymbol => termSymbol -> termSymbol.get)
-      .filter { case (_, value) =>
+      .map(fieldMirror => fieldMirror -> fieldMirror.get)
+      .filter { case (fieldMirror, value) =>
+        def containsChildren: Boolean = fieldMirror.symbol.typeSignature.typeArgs.head <:< treeType
```
Mats-SX: What does `<:<` do here?

okapi-trees/src/test/scala/org/opencypher/okapi/trees/TreeNodeTest.scala
```diff
@@ -244,4 +235,53 @@ class TreeNodeTest extends FunSpec with Matchers {
...
+      NonEmptyList.of(SimpleA(), SimpleA()),
+      List("A", "B"),
+      List(SimpleB(), SimpleB()),
+      Some(1L),
+      None,
+      Some(SimpleD())
+    )
+
+    instance.children.toList should equal(List(
+      SimpleA(), SimpleA(), SimpleB(), SimpleB(), SimpleD()
```
Mats-SX: This is pretty sweet!

okapi-trees/src/test/scala/org/opencypher/okapi/trees/TreeNodeTest.scala
```diff
@@ -244,4 +235,53 @@ class TreeNodeTest extends FunSpec with Matchers {
...
+    ))
+
+    instance.withNewChildren(Array(
+      SimpleA(), SimpleC()
```
Mats-SX: Is it only allowed to do this if there's only one possible place to put these values? So if `Multi` had two fields that had `SimpleA`s as elements (either directly or as type of a list/option) this would throw, as we wouldn't know where to put the one `SimpleA` provided? I suspect there might be a test for this above, but didn't find one when looking briefly.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala
```diff
@@ -301,15 +303,15 @@ object RelationalPlanner {
       if (valueIntos.isEmpty) op else relational.AddInto(op, valueIntos.toList)
     }
 
-    def drop[E <: Expr](expressions: Set[E]): RelationalOperator[T] = {
+    def dropExprSet[E <: Expr](expressions: Set[E]): RelationalOperator[T] = {
```
Mats-SX: Renamed to avoid overloading?

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/VarLengthExpandPlanner.scala
```diff
@@ -37,11 +37,13 @@ import org.opencypher.okapi.relational.impl.planning.RelationalPlanner.{process,
...
 import org.opencypher.okapi.relational.impl.{operators => relational}
 
+import scala.reflect.runtime.universe.TypeTag
+
 trait ExpandDirection
 case object Outbound extends ExpandDirection
 case object Inbound extends ExpandDirection
 
-trait VarLengthExpandPlanner[T <: Table[T]] {
+abstract class VarLengthExpandPlanner[T <: Table[T] : TypeTag] {
```
Mats-SX: Are `trait`s not allowed to have type tags?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSSessionImpl.scala
```diff
@@ -52,7 +52,7 @@ import org.opencypher.spark.api.CAPSSession
 import org.opencypher.spark.impl.CAPSConverters._
 import org.opencypher.spark.impl.table.SparkTable.DataFrameTable
 
-sealed class CAPSSessionImpl(val sparkSession: SparkSession) extends CAPSSession with Serializable {
+sealed class CAPSSessionImpl(val sparkSession: SparkSession) extends CAPSSession {
```
Mats-SX: Why was `Serializable` removed?

okapi-trees/src/main/scala/org/opencypher/okapi/trees/AbstractTreeNode.scala
pstutz: This is just to remove the dollar sign characters from the class names names of objects to get a prettier error message.

okapi-trees/src/main/scala/org/opencypher/okapi/trees/TreeNode.scala
pstutz: It checks if the inner type of the list/option is the same as the type of the tree, i.e. if it's a child node. Child nodes are not considered arguments when printing the tree (they get printed hierarchically below the parent).

okapi-trees/src/test/scala/org/opencypher/okapi/trees/TreeNodeTest.scala
pstutz: It assigns children by going over the arguments from left to right and assigns children with suitable types by also going over the new children from left to right. If anything is left over or a required child cannot be found, then it fails. I should add tests for both failure cases.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala
pstutz: Yes, the overloading caused problems.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/VarLengthExpandPlanner.scala
pstutz: Yes, they need at least an abstract class.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSSessionImpl.scala
pstutz: Is a Spark session serialisable? I assumed it's not.

okapi-trees/src/test/scala/org/opencypher/okapi/trees/TreeNodeTest.scala
pstutz: I just added documentation for how new children assigned, better error messages when it fails, and tests for the errors.

okapi-trees/src/main/scala/org/opencypher/okapi/trees/AbstractTreeNode.scala
Mats-SX: Ohh, it filters the string, I get it now

okapi-trees/src/main/scala/org/opencypher/okapi/trees/TreeNode.scala
Mats-SX: So `<:<` means `sameTypeAs`?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSSessionImpl.scala
Mats-SX: Yes it is: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L82

Most of the time things need to be serialisable in order for the classes to be distributed to all worker JVMs in the Spark cluster.

okapi-trees/src/main/scala/org/opencypher/okapi/trees/TreeNode.scala
pstutz: It says:
```scala
/** Does this type conform to given type argument `that`? */
    def <:< (that: Type): Boolean
```

I'm not 100% sure what that means, but I assume that it's true iff the left hand side is the same or a subtype of the right hand side.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSSessionImpl.scala
pstutz: Interesting, I didn't know that. Is our session actually serialised anywhere? I figured that modified catalog entries might be lost etc. and that it would generally be a mistake to do it.

Mats-SX: All code that enters a Spark job must be serialised, and since we pass the session essentially everywhere it must also be serialisable -- but I'm wondering why this isn't an issue in our tests...

I created a PR for restoring it: https://github.com/opencypher/cypher-for-apache-spark/pull/620 but we should probably test this on a real cluster.

pstutz: Thanks for the PR. I would like to understand any issues arising due to lack serialisation, would be nice if we had a test for them.

PR #621, opened 2018-08-22
Remove custom FS parameter for FS-based PGDS

pstutz:
Also get rid of CAPSFilesystem interface

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/csv/HdfsCsvDataSourceAcceptance.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/csv/HdfsCsvDataSourceAcceptance.scala
@@ -26,6 +26,7 @@
  */
 package org.opencypher.spark.api.io.csv
 
+import org.apache.hadoop.conf.Configuration
 import org.opencypher.okapi.relational.api.graph.RelationalCypherGraph
 import org.opencypher.spark.api.GraphSources
 import org.opencypher.spark.api.io.fs.hdfs.HdfsDataSourceAcceptance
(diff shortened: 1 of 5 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/FSGraphSource.scala
```diff
@@ -58,8 +57,7 @@ class FSGraphSource(
 
   import directoryStructure._
 
-  protected lazy val fileSystem: CAPSFileSystem = customFileSystem.getOrElse(
-    FileSystem.get(caps.sparkSession.sparkContext.hadoopConfiguration))
+  protected lazy val fileSystem: FileSystem = FileSystem.get(caps.sparkSession.sparkContext.hadoopConfiguration)
```
Mats-SX: This is the place where I think we should first consider the `rootPath` as a URI and if it contains a scheme, we should call into a different `FileSystem.get()` signature; I was looking at `FileSystem.get(URI, Configuration)`. That way one could create an `FSGraphSource` which reads from a different filesystem than the default.

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/csv/HdfsCsvDataSourceAcceptance.scala
```diff
@@ -34,8 +35,16 @@ import org.opencypher.spark.impl.table.SparkTable.DataFrameTable
 
 class HdfsCsvDataSourceAcceptance extends HdfsDataSourceAcceptance {
 
+  // Set an incompatible default filesystem to ensure that picking the right filesystem based on the protocol works
+  override def clusterConfig: Configuration = {
```
Mats-SX: I guess it would've been nicer if there was a dedicated test for this rather than using unrelated tests to test this particularity of the interface, but it's OK I think.

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/csv/HdfsCsvDataSourceAcceptance.scala
```diff
@@ -34,8 +35,16 @@ import org.opencypher.spark.impl.table.SparkTable.DataFrameTable
...
+    val cfg = super.clusterConfig
+    val incompatibleDefault = s"s3a://bucket/"
+    cfg.set("fs.defaultFS", incompatibleDefault)
```
Mats-SX: I guess it would've been nicer if there was a dedicated test for this rather than using unrelated tests to test this particularity of the interface, but it's OK I think.

PR #623, opened 2018-08-24
Add schema patterns (node edge node) to Schema

DarthMax:
Schema patterns are either computed (all possible combinations) or provided explicitly.
If any schema patterns are explicitly defined we assume those describe the schema in full. 

An alternative approach could be to make this distinction on a relationship type basis, i.e. for every rel type that occurs in an explicitly defined schema pattern, we make the above assumption and for any other rel types we assume all possible combinations.

### DIFF
--- a/okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
+++ b/okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
@@ -366,6 +366,14 @@ final case class SchemaImpl(
         builder.append(s"no relationship types$EOL")
       }
 
+      if (explicitSchemaPatterns.nonEmpty) {
+        builder.append(s"Explicit schema patterns {$EOL")
+        explicitSchemaPatterns.foreach( p =>
+          builder.append(s"\t$p$EOL")
+        )
+        builder.append(s"}$EOL")
+      }
+
       builder.toString
     }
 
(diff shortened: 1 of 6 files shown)


okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
```diff
@@ -54,11 +55,13 @@ object SchemaImpl {
     schema => Js.Obj(
       VERSION -> Js.Num(1),
```
pstutz: Intentionally leaving schema version at `1` until the release? Makes sense.

okapi-api/src/main/scala/org/opencypher/okapi/impl/schema/SchemaImpl.scala
```diff
@@ -176,6 +192,18 @@ final case class SchemaImpl(
 
   override def relationshipKeys(typ: String): PropertyKeys = relTypePropertyMap.properties(typ)
 
+  override def schemaPatternsFor(
+    knownSourceLabels: Set[String],
+    knownRelTypes: Set[String],
+    knownTargetLabels: Set[String]
+  ): Set[SchemaPattern] = {
+    val possibleSourcePatterns = schemaPatterns.filter(p => knownSourceLabels.subsetOf(p.sourceLabels))
+    val possibleRelTypePatterns = possibleSourcePatterns.filter(p => knownRelTypes.contains(p.relType) || knownRelTypes.isEmpty)
```
pstutz: More efficient to handle empty case with an `if`

pstutz: It would be more efficient to retrieve the relevant patterns by internally storing the schema patterns in a map that uses the relationship type as the key. Given that we don't really use the feature yet, I don't think it matters much. And we could always change the internal representation later on.

PR #637, opened 2018-09-05
Make sure to close all ExecutionContexts or remove them altogether

DarthMax:
Kudos to @jjaderberg and @lutovich


okapi-neo4j-io/src/main/scala/org/opencypher/okapi/neo4j/io/EntityWriter.scala
```diff
@@ -27,9 +27,9 @@
 package org.opencypher.okapi.neo4j.io
 
 import org.apache.logging.log4j.scala.Logging
-import org.neo4j.driver.internal.value.{ListValue, MapValue}
+import org.neo4j.driver.internal.value.ListValue
```
jjaderberg: Can we get rid of the other use(s) of ListValue?

PR #640, opened 2018-09-07
Yield better error message on unimplemented functions

### LINKED ISSUE #632: Improve error message when calling `tail()` on an empty list
trying to apply `tail()` method on a list ,the list is created with `collect()` ,i get the following error  : 

     Exception in thread "main" scala.MatchError: Tail (of class org.opencypher.v9_0.expressions.functions.Tail$)
	at org.opencypher.okapi.ir.impl.typer.SchemaTyper$BasicSignatureBasedTyper$.generateSignaturesFor(SchemaTyper.scala:480)
	at org.opencypher.okapi.ir.impl.typer.SchemaTyper$SignatureBasedInvocationTyper$class.selectSignaturesFor(SchemaTyper.scala:454)
	at org.opencypher.okapi.ir.impl.typer.SchemaTyper$BasicSignatureBasedTyper$.selectSignaturesFor(SchemaTyper.scala:475)
	at org.opencypher.okapi.ir.impl.typer.SchemaTyper$SignatureBasedInvocationTyper$$anonfun$apply$53$$anonfun$apply$55$$anonfun$apply$56.apply(SchemaTyper.scala:438)
	at org.opencypher.okapi.ir.impl.typer.SchemaTyper$SignatureBasedInvocationTyper$$anonfun$apply$53$$anonfun$apply$55$$anonfun$apply$56.apply(SchemaTyper.scala:437)
	at org.atnos.eff.EffImplicits$$anon$1$$anonfun$flatMap$1.apply(Eff.scala:181)
	at org.atnos.eff.EffImplicits$$anon$1$$anonfun$flatMap$1.apply(Eff.scala:181)
         .....
these is my code :
    
      FROM GRAPH nodes
      match(m)
      FROM GRAPH session.inputGraph
       match(n) where n.geohash=m.geohash
       with  collect(n.timehash)as timehashList
       WITH tail(timehashList)  AS result
       construct
       create(h:hashnode{timehashList:result)
     RETURN GRAPH

Mats-SX:
Resolves #632


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/FunctionsBehaviour.scala
```diff
@@ -894,4 +895,12 @@ class FunctionsBehaviour extends CAPSTestSuite with DefaultGraphInit {
       ))
     }
   }
+
+  describe("negative tests") {
+    it("should give a good error message on unimplemented functions") {
+      the [NotImplementedException] thrownBy {
+        caps.cypher("RETURN tail([1, 2])")
+      } should have message "The expression FunctionInvocation(Namespace(List()),FunctionName(tail),false,Vector(Parameter(  AUTOLIST0,List<Any>))) [line 1, column 8 (offset: 7)] is not supported by the system"
```
s1ck: Very nice output. Maybe not super intuitive for the user, but perfect for bug reports.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -519,6 +519,9 @@ object SchemaTyper {
         case f: TypeSignatures =>
           val set = f.signatures.flatMap(_.convert).toSet
           pure(set)
+
+        case _ =>
+          wrong[R, TyperError](UnsupportedExpr(expr)) >> pure(Set.empty)
```
s1ck: Why don't you throw a `NotImplementedException` directly with the appropriate text instead of wrapping it later?

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/FunctionsBehaviour.scala
Mats-SX: I thought about adding a pretty-printer for these AST nodes, but it seemed like a pretty big task.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
Mats-SX: To follow the monadic pattern. We never throw from inside the `SchemaTyper`, instead we collect 'effects' that are only inspected once the whole query has been typed. Some of those effects are errors, which we throw as exceptions.

s1ck: Makes sense. Thanks for the explanation.

PR #642, opened 2018-09-07
Remove cluster profile

Mats-SX:
Simplifies the pom by lifting module listing to the default scope and
removing the special `cluster` profile which had a different module list.
This will now build all the present modules by default.

License listing is disabled for the `spark-cypher-examples` module.
Shade plugin is moved to the modules where it is used.

Co-authored-by: Max Kießling <max.kiessling@neotechnology.com>


s1ck:
closing because of wrong TC links

PR #646, opened 2018-09-17
Improve Neo4jSync documentation and add tests

pstutz:


spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4jSync.scala
```diff
@@ -49,6 +49,8 @@ import scala.concurrent.{Await, Future}
   * Describes sets of properties that uniquely identify nodes/relationships with a given
   * label/relationship type.
   *
+  * If an entity has multiple labels, then it needs to have the entity keys of all its labels.
```
Mats-SX: Is this true? I'd have thought that just one key would be necessary, having more seems superfluous.

Mats-SX: In fact, the test `can do basic Neo4j syncing with merges` above is merging a node `(e:N:M {id: 2})` by just specifying a key for `:N`.

pstutz: In the mentioned case I think the key is `id` for both labels.

pstutz: We wrote tests and ran into problems when there were multiple labels and the entity key for one of them was missing. I think we could be more flexible and maybe just pick one of the entity keys that is present on the label combination (according to the schema). Or we could allow users to explicitly define entity keys for label combinations, but that might have interactions with index creation.

pstutz: AFAIK we can only define indexes on single labels. This means that allowing label combinations to not have all the entity keys for all the individual labels interacts badly with our ability to create indexes.

Mats-SX: Exactly, that's why we opted to use just a single label for `EntityKeys`.

You mention you ran into problems, but at the same time we have the test above that I mentioned. So it seems that only in some cases is it an issue if there's a key missing for a label which is part of a combination. Of course, if the labels in the label combination appears singularly on nodes, we need a key for all of them, like in a graph of this kind:

```
(:A) // requires key for A
(:A:B) // requires key for A or B
(:B) // requires key for B
```

but I'm still not convinced this is true in this case:

```
(:A:B) // requires key for A or B
(:A) // requires key for A
```

ie, no need for key on B in the below graph provided that a key using just `:A` is possible.

PR #647, opened 2018-09-17
Remove schema procedure

pstutz:
Still requires comprehensive type name conversions and testing. There seem to be some corner cases such as values in a list where we should always assume nullability: The `okapi.schema` procedure only has a single nullability flag for the list itself.


pstutz:
TODO: Adapt wiki entry for schema procedure.

pstutz:
## Status
Has three issues:
- **Nullability bug** reproduced with
```
test("input order does not affect schema") {
    val schema = schemaFor(
      """|CREATE (b1:B { type: 'B1' })
         |CREATE (b2:B { type: 'B2', size: 5 })""".stripMargin)
    schema should equal(Schema.empty
      .withNodePropertyKeys("B")("type" -> CTString, "size" -> CTInteger.nullable)
    )
  }
```
@SaschaPeukert kindly agreed to provide a fix for Neo4j 3.5 release and back port it to 3.4. Thanks a lot!
- **Complex non-deterministic nullability bug**: Reproduced by not completely resetting the Neo4j DB in `Neo4jSyncTest` (new name `Neo4jGraphMergeTest`): https://github.com/opencypher/cypher-for-apache-spark/pull/647/files#diff-842fdd31ba65fa147e9b742ecfa06718R58
We hope that it is caused by the same bug as the first issue.
- The conversion of Cypher type strings returned by the procedure are not fully tested yet (and probably the types are not fully covered). Ideally we use the list of all possible `CypherType` values, for each write them to Neo4j, and then ensure that the string returned by `okapi.schema()` is correctly parsed into the same value. Some negative tests for types such as `Time` should be added as well.

pstutz:
[Drop Okapi Schema Procedure and replace with Neo4j Procedure (Neo4j versions!)](https://trello.com/c/ehWTCanm/200-drop-okapi-schema-procedure-and-replace-with-neo4j-procedure-neo4j-versions)

PR #648, opened 2018-09-18
Make entire graph name unconfigurable

s1ck:


s1ck:
note: reopened to check if PR build links are being updated

PR #653, opened 2018-09-20
Make EntityTable showable

Mats-SX:
CAPSRelationshipTable is a CAPSRecords and implements RecordsBehaviour, but it did not support `.collect()` or `.show()` properly because there was no relationship type information in the computer header. This is instead recorded on the variable's type, which this PR now passes on as a fallback when no type is found in the header.

Also cleans up a test class and provides output capture and validation in `BaseTestSuite` (good for testing `.show()`)


okapi-testing/src/main/scala/org/opencypher/okapi/testing/BaseTestSuite.scala
```diff
@@ -61,4 +64,14 @@ abstract class BaseTestSuite extends FunSpec with Matchers with MockitoSugar {
...
   def test(name: String, tags: Tag*)(testFun: => Any /* Assertion */ )(implicit pos: source.Position): Unit =
     it(name, tags: _*)(testFun)
+
+  val outCapture = new ByteArrayOutputStream()
+  val printer = new PrintStream(outCapture, true, "UTF-8")
+  val capturingPrintOptions: PrintOptions = {
+    PrintOptions(printer)
+  }
+
+  def assertPrinted(expectation: String): Assertion = {
```
s1ck: nice to have this generally available

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/io/EntityTable.scala
```diff
@@ -89,6 +90,13 @@ trait EntityTable[T <: Table[T]] extends RelationalCypherRecords[T] {
   }
 }
 
+object EntityTable {
+  /**
+    * Default variable name to use for Scan operators for RelationalCypherGraphs, as well as EntityTable entities.
```
s1ck: We should use scala doc to refer to `RelationalCypherGraph#scanOperator`

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
```diff
@@ -48,36 +49,89 @@ class CAPSRecordsAcceptanceTest extends CAPSTestSuite with CAPSNeo4jServerFixtur
...
+    val table = sparkSession.createDataFrame(Seq(ActedIn(1, 2, 3), ActedIn(2, 3, 4)))
+    val relTable: CAPSRelationshipTable = CAPSRelationshipTable.fromMapping(mapping, table)
+
+    relTable.collect.toBag should equal(Bag(
+      CypherMap("entity" -> CAPSRelationship(1, 2, 3, "ACTED_IN")),
+      CypherMap("entity" -> CAPSRelationship(2, 3, 4, "ACTED_IN"))
+    ))
+  }
+
+  it("CAPSRelTable can show") {
```
s1ck: We should add a test for NodeTable, too

s1ck: Done, this revealed an issue with show on `CAPSNodeTable` missing implied labels in the output. This should be fixed now.

s1ck:
@Mats-SX and I discussed this PR and decided to not merge it now as it only adds hacks around our current, partially inconsistent handling of node labels / rel types on vars (logical) and physical columns. A better approach would be to define a way to consistently access schema information, e.g. by attaching a graph schema to a `CypherType`.

PR #663, opened 2018-09-30
fixes #650: Feature request: Support `RegexMatch`

### LINKED ISSUE #650: Feature request: Support `RegexMatch`
Trying to effect some `slice` and `dice` operations in the Graph with Regex: 

     session.cypher("""
                                      FROM GRAPH inputGraph
                                     |  MATCH (from)-[via]->(to)
                                     |  FROM GRAPH nodes
                                     |  MATCH (n), (m)
                                     |  WHERE from.geohash =~ 'dr5*'
                                     |  CONSTRUCT
                                     |    CREATE (n)-[COPY OF via]->(m)
                                     |  RETURN GRAPH
                                  """.stripMargin).graph
    
i get this exceptions : 

    Exception in thread "main" org.opencypher.okapi.impl.exception.NotImplementedException: Not yet able to convert expression: RegexMatch(Property(Variable(from),PropertyKeyName(geohash)),Parameter(  AUTOSTRING0,String))

conker84:
Hi guys,
I implemented this feature request.

In `CAPSRecordsAcceptanceTest` you'll find two test:

```scala
  it("filter rels on property regular expression") {
    // Given
    val query = """MATCH (a:Actor)-[r:ACTED_IN]->() WHERE r.charactername =~ '(\\w+\\s*)*Du\\w+' RETURN r.charactername"""

    // When
    val result = graph.cypher(query)

    // Then
    val records = result.records.collect
    records.toBag should equal(Bag(CypherMap("r.charactername" -> "Henri Ducard"),
      CypherMap("r.charactername" -> "Albus Dumbledore")))
  }

  it("filter nodes on property regular expression") {
    // Given
    val query = """MATCH (p:Person) WHERE p.name =~ '\\w+ Redgrave' RETURN p.name"""

    // When
    val result = graph.cypher(query)

    // Then
    val records = result.records.collect
    records.toBag should equal(Bag(CypherMap("p.name" -> "Michael Redgrave"),
      CypherMap("p.name" -> "Vanessa Redgrave"),
      CypherMap("p.name" -> "Corin Redgrave"),
      CypherMap("p.name" -> "Jemma Redgrave"),
      CypherMap("p.name" -> "Roy Redgrave")))

  }
```

Under the hood it uses the Spark's `rlike` function:

```scala
      def rlike(lhs: Expr, name: String): Column = {
        val regex: String = parameters(name).unwrap.toString
        lhs.asSparkSQLExpr.rlike(regex)
      }
```

Let me know what do you think about it.
Thanks

Signed-off-by: Andrea Santurbano santand@gmail.com

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
@@ -102,36 +102,6 @@ class CAPSRecordsAcceptanceTest extends CAPSTestSuite with CAPSNeo4jServerFixtur
     result.records shouldHaveSize 3 // andContain "Christopher Nolan"
   }
 
-  it("filter rels on property regular expression") {
-    // Given
-    val query = """MATCH (a:Actor)-[r:ACTED_IN]->() WHERE r.charactername =~ '(\\w+\\s*)*Du\\w+' RETURN r.charactername"""
-
-    // When
-    val result = graph.cypher(query)
-
-    // Then
-    val records = result.records.collect
-    records.toBag should equal(Bag(CypherMap("r.charactername" -> "Henri Ducard"),
-      CypherMap("r.charactername" -> "Albus Dumbledore")))
-  }
-
-  it("filter nodes on property regular expression") {
-    // Given
-    val query = """MATCH (p:Person) WHERE p.name =~ '\\w+ Redgrave' RETURN p.name"""
-
-    // When
-    val result = graph.cypher(query)
-
-    // Then
-    val records = result.records.collect
-    records.toBag should equal(Bag(CypherMap("p.name" -> "Michael Redgrave"),
-      CypherMap("p.name" -> "Vanessa Redgrave"),
-      CypherMap("p.name" -> "Corin Redgrave"),
-      CypherMap("p.name" -> "Jemma Redgrave"),
-      CypherMap("p.name" -> "Roy Redgrave")))
-
-  }
-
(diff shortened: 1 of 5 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -158,6 +163,10 @@ object SparkSQLExprMapper {
         case Contains(lhs, rhs) =>
           lhs.asSparkSQLExpr.contains(rhs.asSparkSQLExpr)
 
+        case RegexMatch(prop, Param(name)) => rlike(prop, name)
+        case RegexMatch(Param(name), prop) => rlike(prop, name)
```
DarthMax: I don't think this case is possible, the front end will pass the parameters in a specific order to the AST

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
```diff
@@ -101,6 +102,36 @@ class CAPSRecordsAcceptanceTest extends CAPSTestSuite with CAPSNeo4jServerFixtur
     result.records shouldHaveSize 3 // andContain "Christopher Nolan"
   }
 
+  it("filter rels on property regular expression") {
```
DarthMax: Can you move these tests into `org.opencypher.spark.impl.acceptance.ExpressionBehaviour`?

conker84:
Hi @DarthMax, thanks for the review I implemented your suggestions.

conker84:
@DarthMax ping

PR #665, opened 2018-10-02
Add test for Neo4jGraphMergeBug

Mats-SX:
The test is very raw and can almost guaranteed be minimised to drill down into the error


soerenreichardt:
This was only an error in the usage of Neo4j graph merge rather than the program logic.

PR #669, opened 2018-10-09
implement pi function

soerenreichardt:


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -751,6 +751,11 @@ final case class E()(var cypherType: CypherType = CTWildcard) extends NullaryFun
   override def withCypherType(ct: CypherType): E = copy()(ct)
 }
 
+final case class Pi()(var cypherType: CypherType = CTWildcard) extends NullaryFunctionExpr {
```
Mats-SX: I don't think this should be a `var`, but I can see a lot of other functions around here use `var` -- is this a case of copy/paste that's gone wrong or do we rely on `var`s here?

Generally I would suggest it should be `val` in this and all other classes here.

PR #670, opened 2018-10-09
Add DDL Parser

pstutz:


okapi-api/src/main/scala/org/opencypher/okapi/api/schema/Schema.scala
```diff
@@ -120,80 +136,56 @@ trait Schema {
...
-  def nodeKeys(labels: Set[String]): PropertyKeys
-
-  /**
-    * Given a set of labels that a node definitely has, returns its property schema.
+    * Given a set of labels that a node definitely has, returns its property keys and their types.
     */
-  def nodeKeys(labels: String*): PropertyKeys = {
-    nodeKeys(labels.toSet)
-  }
+  def nodePropertyKeys(labels: Set[String]): PropertyKeys
```
Mats-SX: Why was the convenience varargs method removed?

okapi-api/src/test/scala/org/opencypher/okapi/api/schema/SchemaTest.scala
```diff
@@ -643,4 +691,57 @@ class SchemaTest extends FunSpec with Matchers {
...
+        Schema.empty.withNodeKey("A", Set.empty)
+      }
+      a[SchemaException] shouldBe thrownBy {
+        Schema.empty.withRelationshipKey("A", Set.empty)
+      }
+    }
+
+    it("merges overlapping keys during schema merge") {
+      val schema1 = Schema.empty.withNodePropertyKeys("A")().withNodeKey("A", Set("foo"))
+      val schema2 = Schema.empty.withNodePropertyKeys("A")().withNodeKey("A", Set("bar"))
```
Mats-SX: It seems to me that the set of properties used for the key should be a subset of all the properties for the label, and otherwise this should fail.

```
Schema.empty.withNodePropertyKeys("A")("foo" -> CTBoolean, "antoher" -> CTString).withNodeKey("A", Set("foo")) // OK because 'foo' is a property of :A
```

okapi-api/src/test/scala/org/opencypher/okapi/api/schema/SchemaTest.scala
```diff
@@ -643,4 +691,57 @@ class SchemaTest extends FunSpec with Matchers {
...
+      val joinedSchema = schema1 ++ schema2
+
+      joinedSchema.nodeKeys shouldEqual Map("A" -> Set("foo", "bar"))
```
Mats-SX: This seems correct, given the modification above.

okapi-api/src/main/scala/org/opencypher/okapi/api/schema/Schema.scala
s1ck: We removed methods that were only used in tests. This was one of them.

okapi-api/src/test/scala/org/opencypher/okapi/api/schema/SchemaTest.scala
```diff
@@ -643,4 +691,57 @@ class SchemaTest extends FunSpec with Matchers {
     }
   }
 
+  describe("entity keys") {
+    it("adds node keys") {
+      val schema = Schema.empty
+        .withNodePropertyKeys("A")()
+        .withNodeKey("A", Set("foo", "bar"))
+      schema.nodeKeys shouldEqual Map("A" -> Set("foo", "bar"))
```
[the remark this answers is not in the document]
s1ck: It does now :)

okapi-api/src/main/scala/org/opencypher/okapi/api/schema/Schema.scala
Mats-SX: That's a good rationale.

okapi-api/src/test/scala/org/opencypher/okapi/api/schema/SchemaTest.scala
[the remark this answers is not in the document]
DarthMax: Yay

PR #672, opened 2018-10-11
Timestamp support

soerenreichardt:


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/functions/TimeStamp.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import org.opencypher.v9_0.expressions._
+import org.opencypher.v9_0.expressions.functions.Function
+import org.opencypher.v9_0.util.symbols._
+
+case object TimeStamp extends Function with TypeSignatures {
+  def name = "timestamp"
+
+  override val signatures = Vector(
+    TypeSignature(argumentTypes = Vector(), outputType = CTMap)
```
Mats-SX: Output type should be a `CTInteger`

PR #679, opened 2018-10-16
Clean up CAPSRecordsAcceptanceTest

### LINKED ISSUE #504: Re-enable containment tests in CAPSRecordsAcceptanceTest
contain syntax does not work for maps -> https://github.com/scalatest/scalatest/issues/1224

Maybe we can find a way to work around that?

Mats-SX:
Commits cherry-picked from #653 

Fixes #504

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
@@ -49,20 +49,28 @@ class CAPSRecordsAcceptanceTest extends CAPSTestSuite with CAPSNeo4jServerFixtur
   private lazy val graph: RelationalCypherGraph[DataFrameTable] =
     GraphSources.cypher.neo4j(neo4jConfig).graph(entireGraphName).asCaps
 
-  it("convert to CypherMaps") {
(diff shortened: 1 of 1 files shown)


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/CAPSRecordsAcceptanceTest.scala
```diff
@@ -49,36 +46,46 @@ class CAPSRecordsAcceptanceTest extends CAPSTestSuite with CAPSNeo4jServerFixtur
...
   }
 
   it("label scan and project") {
     // When
     val result = graph.cypher("MATCH (a:Person) RETURN a.name")
 
     // Then
-    result.records shouldHaveSize 15 // andContain "Rachel Kempson"
+    result.records.size shouldBe 15
+    result.records.collect should contain(CypherMap("a.name" -> "Rachel Kempson"))
```
s1ck: Did you try the `andContain` syntax? We had to comment that at some point because of an issue with the `Bag` implementation. Maybe it works now?

Mats-SX: No I just removed the `andContain` in favour of standard assertions from ScalaTest

PR #693, opened 2018-10-29
Add AST rewriter for label filter pushdown

### LINKED ISSUE #513: Entity variables in planners have the wrong CypherType
The AST removes the labels from the pattern and moves them into to `WHERE` clauses. They are later added back in the logical optimizer step. However, typing of the pattern and thus typing of the entity variables already happens at IR planning time. The variables and their CypherTypes are then used as input for subsequent planning phases. This leads to a mismatch of the variables stored in the RecordHeader and the variables references in the operators

DarthMax:
Fixes #513 
Co-authored-by: Sören Reichardt <soren.reichardt@neotechnology.com>

### DIFF
--- a/okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/parse/rewriter/pushLabelsIntoScansTest.scala
+++ b/okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/parse/rewriter/pushLabelsIntoScansTest.scala
@@ -1,3 +1,29 @@
+/*
+ * Copyright (c) 2016-2018 "Neo4j Sweden, AB" [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Attribution Notice under the terms of the Apache License 2.0
+ *
+ * This work was created by the collective efforts of the openCypher community.
+ * Without limiting the terms of Section 6, any Derivative Work that is not
+ * approved by the public consensus process of the openCypher Implementers Group
+ * should not be described as “Cypher” (and Cypher® is a registered trademark of
+ * Neo4j Inc.) or as "openCypher". Extensions by implementers or prototypes or
+ * proposals for change that have been documented or implemented should only be
+ * described as "implementation extensions to Cypher" or as "proposed changes to
+ * Cypher that are not yet approved by the openCypher community".
+ */
 package org.opencypher.okapi.ir.impl.parse.rewriter
 
 import org.opencypher.okapi.ir.test.support.RewriterTestSupport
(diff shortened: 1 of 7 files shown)


okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/parse/rewriter/pushLabelsIntoScansTest.scala
```diff
@@ -0,0 +1,138 @@
...
+      """MATCH (n:Foo)
+        |WHERE n:Bar
+        |RETURN n
+      """.stripMargin,
+      """MATCH (n:Foo:Bar)
+        |RETURN n
+      """.stripMargin)
+  }
+
+  test("push complex where predicate labels into pattern") {
```
Mats-SX: Let's maybe add a test where the `HasLabel` is composed with another type of predicate as well:

```
MATCH (n) WHERE n.p = 5 AND n:Foo
```

okapi-logical/src/test/scala/org/opencypher/okapi/logical/impl/logical/LogicalPlannerTest.scala
```diff
@@ -108,150 +108,140 @@ class LogicalPlannerTest extends BaseTestSuite with IrConstruction {
...
+//          HasLabel(Var("a")(CTNode), Label("Administrator"))(CTBoolean),
+//          HasLabel(Var("g")(CTNode), Label("Group"))(CTBoolean),
+//          Equals(Property(Var("g")(CTNode(Set("Group"))), PropertyKey("name"))(CTString), Param("foo")(CTString))(
+//            CTBoolean)
+//        )
+//      )
+//    )
+//
+//    result should equalWithoutResult(expected)
+//  }
```
Mats-SX: Either delete or uncomment and fix this. I'm not really comfortable just deleting it as it is the only tests we have that test the pipeline at this level. It's pretty tedious to write though...

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOptimizer.scala
```diff
@@ -26,35 +26,24 @@
...
       discardScansForNonexistentLabels,
       replaceCartesianWithValueJoin)
-    optimizationRules.foldLeft(input) {
+      optimizationRules.foldLeft(input) {
       // TODO: Evaluate if multiple rewriters could be fused
       case (tree: LogicalOperator, optimizationRule) => BottomUp[LogicalOperator](optimizationRule).transform(tree)
     }
   }
 
-  def labelsForVariables(op: LogicalOperator): Map[Var, Set[String]] = {
```
Mats-SX: Nice replacement

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOperatorProducer.scala
```diff
@@ -82,7 +82,12 @@ class LogicalOperatorProducer {
   }
 
   def planNodeScan(node: IRField, prev: LogicalOperator): NodeScan = {
-    NodeScan(node, prev, prev.solved.withField(node))
+    val solvedWithField = prev.solved.withField(node)
+    val solvedWithPredicates = node.cypherType match {
+      case CTNode(labels, _) => solvedWithField.withPredicates(labels.map(l => HasLabel(node, Label(l))(CTBoolean)).toSeq:_*)
+      case _ => solvedWithField
```
Mats-SX: Isn't this some kind of error?

okapi-logical/src/test/scala/org/opencypher/okapi/logical/impl/logical/LogicalPlannerTest.scala
DarthMax: We have fixed them for now. However, we should think of a better way to do these tests

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/parse/rewriter/pushLabelsIntoScansTest.scala
DarthMax: :+1:

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOperatorProducer.scala
DarthMax: Not if we assume that the CypherType will always be CTNode in which case the other case just covers the case without labels

PR #708, opened 2018-11-07
Add Graph DDL generation and integration demo for LDBC

s1ck:


spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/LdbcHiveExample.scala
```diff
@@ -0,0 +1,101 @@
...
+import org.opencypher.okapi.api.graph.Namespace
+import org.opencypher.spark.api.{CAPSSession, GraphSources}
+import org.opencypher.spark.testing.utils.FileSystemUtils._
+import org.opencypher.spark.util.LdbcUtil._
+import org.opencypher.spark.util.{ConsoleApp, LdbcUtil}
+/**
+  * This demo reads data generated by the LDBC SNB data generator and performs the following steps:
+  *
+  * 1) Loads the raw CSV files into Hive tables
+  * 2) Normalized tables according to the LDBC schema (i.e. place -> [City, Country, Continent]
```
tobias-johansson:
```suggestion
  * 2) Normalizes tables according to the LDBC schema (i.e. place -> [City, Country, Continent]
```

spark-cypher-examples/src/main/resources/ldbc/ddl/data-sources.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "storageFormat": "hive",
+    "dataSourceName": "hive"
```
tobias-johansson: Perhaps using another name instead of `hive` makes it easier to understand that this is the identifier we are referring to in the `SET SCHEMA`?

PR #713, opened 2018-11-12
Add hash-based id generation to SQL PGDS

s1ck:


spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -210,6 +238,43 @@ case class SqlPropertyGraphDataSource(
...
+    idColumnNames: List[String],
+    newIdColumn: String
+  ): DataFrame = {
+    val viewLiteral = functions.lit(elementViewKey.qualifiedViewId.view)
+    val nodeTypeLiterals = elementViewKey.elementType.toSeq.sorted.map(functions.lit)
+    val idColumns = idColumnNames.map(dataFrame.col)
+    dataFrame.withHashColumn(Seq(viewLiteral) ++ nodeTypeLiterals ++ idColumns, newIdColumn)
+  }
+
+  private def withHashId[T <: ElementViewKey](
```
tobias-johansson: Perhaps the two `withHashId` methods could get better names and some scala-doc?
Could act like an explanation of the hash-based stategy in general

PR #719, opened 2018-11-20
Allow usage of schema defined element keys in Neo4jGraph merge

### LINKED ISSUE #682: Replace separate entity keys with the ones defined on the schema in Neo4jGraphMerge
Maybe we could do `PropertyGraph.withSchema(...)`? 
This would also force us to consider would happen if one where to drop a label, for instance?
Perhaps `PropertyGraph.withNodeKeys(...)`/`PropertyGraph.withRelationshipKeys(...)` would offer a more manageable interface?

DarthMax:
Fixes #682


spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/Neo4jMergeExample.scala
```diff
@@ -55,10 +56,8 @@ object Neo4jMergeExample extends ConsoleApp {
   ).withSchemaProcedure
 
   // Define the node and relationship keys
-  val entityKeys = EntityKeys(
-    Map("Person" -> Set("name")),
-    Map("FRIEND_OF" -> Set("id"), "MARRIED_TO" -> Set("id"))
-  )
+  val nodeKeys = Map("Person" -> Set("name"))
+  val relKeys = Map("FRIEND_OF" -> Set("id"), "MARRIED_TO" -> Set("id"))
```
Mats-SX: Wouldn't one want to add these to the schema directly somehow? Perhaps we ought to allow people to specify keys on a `PropertyGraph`? In this example where the graph is defined using Cypher `CONSTRUCT` we can't pass it into the definition. I guess we can do that for Graph DDL-defined graphs though?

spark-cypher-examples/src/main/scala/org/opencypher/spark/integration/Customer360IntegrationDemo.scala
```diff
@@ -126,21 +127,19 @@ object Customer360IntegrationDemo extends ConsoleApp {
...
-      "CustomerRep" -> Set("empNo"),
-      "AccountHolder" -> Set("accountHolderId"),
-      "Policy" -> Set("policyAccountNumber")
-    )
+  val nodeKeys = Map(
+    "Interaction" -> Set("interactionId"),
+    "Customer" -> Set("customerId"),
+    "CustomerRep" -> Set("empNo"),
+    "AccountHolder" -> Set("accountHolderId"),
+    "Policy" -> Set("policyAccountNumber")
```
Mats-SX: In this case the source graph came from Graph DDL (`CONSTRUCT ON`) but I guess we don't have anything that validates whether the `CREATE`s would violate the keys. In this example they aren't creating any nodes so I guess it could in theory propagate the keys to the `CONSTRUCT`ed graph...

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4JGraphMergeTest.scala
```diff
@@ -191,29 +198,22 @@ class Neo4JGraphMergeTest extends CAPSTestSuite with CAPSNeo4jServerFixture with
...
-          "Employee" -> Set("baz")
-        ),
-        Map(
-          "REL" -> Set("a")
-        )
-      )
-
-      Neo4jGraphMerge.createIndexes(neo4jConfig, entityKeys)
+      val nodeKeys = Map("Person" -> Set("name", "bar"), "Employee" -> Set("baz"))
+      val relKeys = Map("R" -> Set("a"))
```
Mats-SX: It was `REL` in the removed code -- is there a bug here?

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4JGraphMergeTest.scala
```diff
@@ -279,18 +279,11 @@ class Neo4JGraphMergeTest extends CAPSTestSuite with CAPSNeo4jServerFixture with
...
-        Map(
-          "Person" -> Set("name", "bar"),
-          "Employee" -> Set("baz")
-        ),
-        Map(
-          "REL" -> Set("a")
-        )
-      )
+      val nodeKeys = Map("Person" -> Set("name", "bar"), "Employee" -> Set("baz"))
+      val relKeys = Map("R" -> Set("a"))
```
Mats-SX: It was `REL` in the removed code

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4jGraphMerge.scala
```diff
@@ -47,65 +47,44 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
-  * If a label always appears in conjunction with another label (e.g. :Employee:Person, where being an employee implies
-  * being a person), then it is enough to only provide an entity key for one of the labels (e.g. :Person).
-  *
-  * Relationship keys are optional, if none are provided, then it is implicitly assumed that there is at most one
-  * relationship with a given type between two nodes.
-  *
-  * @param nodeKeys maps a label to a set of property keys, which uniquely identify a node
-  * @param relKeys  maps a relationship type to a set of property keys, which uniquely identify a relationship
-  */
-case class EntityKeys(
```
Mats-SX: Nice!

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4jGraphMerge.scala
```diff
@@ -196,9 +155,18 @@ object Neo4jGraphMerge extends Logging {
...
-  private def validateEntityKeys(graphSchema: Schema, entityKeys: EntityKeys): Unit = {
-    entityKeys.nodeKeys.foreach { nodeKey => graphSchema.withNodeKey(nodeKey._1, nodeKey._2) }
-    entityKeys.relKeys.foreach { relKey => graphSchema.withRelationshipKey(relKey._1, relKey._2) }
+
+  private def combineEntityKeys(
+    schema: Schema,
+    nodeKeys: Option[NodeKeys],
+    relationshipKeys: Option[RelationshipKeys]
+  ): Schema = {
+    val withNodeKeys = nodeKeys.getOrElse(Map.empty).foldLeft(schema) {
```
Mats-SX:
```suggestion
    val withNodeKeys = nodeKeys.foldLeft(schema) {
```

Mats-SX: I think that works on an `Option` already (doesn't iterate if it's a `None`)

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4jGraphMerge.scala
```diff
@@ -196,9 +155,18 @@ object Neo4jGraphMerge extends Logging {
...
+      case (acc, (label, keys)) => acc.withNodeKey(label, keys)
+    }
+    relationshipKeys.getOrElse(Map.empty).foldLeft(withNodeKeys) {
```
Mats-SX:
```suggestion
    relationshipKeys.foldLeft(withNodeKeys) {
```

Mats-SX:
I wonder if we can add a test where are able to write a graph with `Neo4jGraphMerge` without passing in keys if the graph was defined in Graph DDL (directly grab from SQL PGDS and into `.merge`)

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/Neo4jMergeExample.scala
DarthMax: In the ticket (and I think a wall walk) we discussed that it is not very clear how the user would specify node keys for existing graphs as this would modify the graph (an operation that our system assumes never happens). We thus decided to postpone that decision and allow overriding the element keys.

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4JGraphMergeTest.scala
DarthMax: yes 
```suggestion
      val relKeys = Map("REL" -> Set("a"))
```

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4JGraphMergeTest.scala
DarthMax:
```suggestion
      val relKeys = Map("REL" -> Set("a"))
```

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/Neo4jMergeExample.scala
Mats-SX: I was thinking more like attaching keys to an existing graph and "promising" that the keys are valid... but in this case then users would still define the keys externally, so it's a cosmetic change at best compared to what one does in this modification... you're right let's wait.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/neo4j/sync/Neo4jGraphMerge.scala
DarthMax: :+1:

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/Neo4jMergeExample.scala
DarthMax: that would still be a modification of the graph. This might not be important for this particular use case since the user has to pass a graph object directly. However, imagine that in the future, we use the node keys within cypher queries. The user could not specify node keys on a graph and then use that graph in a `FROM GRAPH` clause.

DarthMax:
I tried but I could not find an elegant way to define a graph with a given schema

spark-cypher-examples/src/main/scala/org/opencypher/spark/examples/Neo4jMergeExample.scala
Mats-SX: How do you mean a modification of the graph? I'm thinking about a case where we just define keys for a graph that we "know" are true -- but perhaps you see this as a modification in some definition?

PR #724, opened 2018-11-28
Disable nullability checks for spark-cypher to improve scan performance

DarthMax:


spark-cypher-examples/src/main/resources/example_outputs/NorthwindJdbcExample.out
```diff
@@ -7,7 +7,7 @@
 Node labels {
 	:Order
 		customerID: STRING?
-		orderID: INTEGER
+		orderID: INTEGER?
```
Mats-SX: Why is it necessary to make all these nullable? Does something break if we have them without `?`?

DarthMax: This is the schema we get from the graph. Which in turn is derived from the dataframes, which do not have non nullable fields any longer

PR #727, opened 2018-12-03
Refactor graph type handling

tobias-johansson:
Existing implementation of type scoping and resolution etc. is very ad hoc. Refactoring to make it less messy

- Stack-like data structure for graph types, with "inheritance"
- Contains the single implementation for 
  - validation
  - reference resolution
  - definition shadowing


graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -74,212 +90,172 @@ object GraphDdl {
...
-          .flatMap(em => toEdgeToViewMappings(okapiSchema, graph.maybeSetSchema, em))
-      )
-    }
+    parent: GraphType,
+    graph: GraphDefinitionWithContext
+  ): Graph = {
+    val parts = GraphParts(graph.definition.statements)
+    val graphType = parent
+      .push(parts.graphTypeStatements)
+      .push(parts.nodeMappings.map(_.nodeType) ++ parts.relMappings.map(_.relType))
```
Mats-SX: Does this mean we infer node and rel types from the mappings? It will then resolve the ETs for these recursively through graph -> graph type -> global?

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -74,212 +90,172 @@ object GraphDdl {
...
-      case (currentSchema, PatternDefinition(sourceLabelCombinations, _, relTypes, _, targetLabelCombinations)) =>
-        val expandedPatterns = for {
-          sourceCombination <- sourceLabelCombinations
-          relType <- relTypes
-          targetLabelCombination <- targetLabelCombinations
-        } yield SchemaPattern(sourceCombination, relType, targetLabelCombination)
-        currentSchema.withSchemaPatterns(expandedPatterns.toSeq: _*)
-    }
+    private def resolveElementType(name: String): ElementTypeDefinition =
+      allElementTypes.getOrElse(name, unresolved(s"Unresolved element type", name))
```
Mats-SX: Should we not mention the possible element types here, in case one isn't found?

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
```diff
@@ -222,7 +222,18 @@ class GraphDdlTest extends FunSpec with Matchers {
       | (Person3, Person4)
       |)
     """.stripMargin)
-    e.getFullMessage should (include("fooSchema") and include("(Person3,Person4)") and include("Person1") and include("Person2"))
+    e.getFullMessage should (include("fooSchema") and include("Person3") and include("Person4"))
```
Mats-SX: It seems like a slight regression that we do not mention the possible ETs here.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -74,212 +90,172 @@ object GraphDdl {
...
-                Err.incompatibleTypes(
-                  s"""|Incompatible property types for property key: $key
-                      |Conflicting types: ${currProps(key)} and ${labelProperties(key)}
-                 """.stripMargin)
-              }
-            }
-            currProps ++ labelProperties
-          }
-          currentSchema.withNodePropertyKeys(labelCombo, comboProperties)
+    lazy val asOkapiSchema: Schema = {
```
s1ck: I very much like how this looks now :)

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
tobias-johansson: Yeah that's right. Each push will resolve the statements at that stage

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
tobias-johansson: Removed that intentionally actually, since the error message is likely to become unwieldy when you have more than a few element types.
We could of course get clever here, but I think this is something we can leave for later. WDYT?

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
tobias-johansson: See above

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -74,212 +90,172 @@ object GraphDdl {
...
 
   def apply(ddl: DdlDefinition): GraphDdl = {
 
-    val ddlParts = toDdlParts(ddl.statements)
+    val ddlParts = DdlParts(ddl.statements)
 
-    val elementTypeDefinitions: Map[String, ElementTypeDefinition] = ddlParts.elementTypes
-      .validateDistinctBy(_.name, "Duplicate element type")
-      .keyBy(_.name)
+    val global = GraphType().push(ddlParts.elementTypes)
```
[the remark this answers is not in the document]
tobias-johansson: I'd like to argue that there should be an empty scope at the root :)
That makes the operation of `push`ing (i.e. resolving), completely the same everywhere - the global things *really are* added to an empty scope, after all.

If we did `GraphType(elementTypes = ddlParts.elementTypes)` we wouldn't get the validation etc. that is present in `push`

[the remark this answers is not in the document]
tobias-johansson: Compare for instance with `org.opencypher.okapi.api.schema.Schema.empty`

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -74,212 +90,172 @@ object GraphDdl {
...
     val graphs = ddlParts.graphs
-      .validateDistinctBy(_.definition.name, "Duplicate graph name")
-      .map { graph => tryWithContext(s"Error in graph ${graph.definition.name}") {
+      .map { graph => tryWithGraph(graph.definition.name) {
         val graphType = graph.definition.maybeGraphTypeName
           .map(name => graphTypes.getOrFail(name, "Unresolved graph type"))
-          .getOrElse(TypeParts())
-        toGraph(graphType, elementTypeDefinitions, graph)
+          .getOrElse(global)
+        toGraph(graphType, graph)
```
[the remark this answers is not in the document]
tobias-johansson: Yeah everything we need to know is captured in the stack of GraphType:s

PR #731, opened 2018-12-04
Ddl edge types as triplets

tobias-johansson:


graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -354,6 +347,15 @@ object GraphDdl {
     }
   }
 
+  private def toNodeType(nodeTypeDefinition: NodeTypeDefinition): NodeType =
+    NodeType(nodeTypeDefinition.elementTypes)
```
Mats-SX: Maybe this should be an `apply()` on the `NodeType` object companion, instead of a private helper?

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -354,6 +347,15 @@ object GraphDdl {
...
+
+  private def toRelType(relTypeDefinition: RelationshipTypeDefinition): RelationshipType =
+    RelationshipType(
+      startNodeType = toNodeType(relTypeDefinition.sourceNodeType),
+      elementType = relTypeDefinition.elementType,
+      endNodeType = toNodeType(relTypeDefinition.targetNodeType))
```
Mats-SX: Same here; perhaps an `apply()` constructor would be a bit clearer.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -371,6 +373,18 @@ object GraphDdl {
...
+  private def tryWithGraphType[T](name: String)(block: => T): T =
+    tryWithContext(s"Error in graph type: $name")(block)
+
+  private def tryWithGraph[T](name: String)(block: => T): T =
+    tryWithContext(s"Error in graph: $name")(block)
+
+  private def tryWithNode[T](labels: Set[String])(block: => T): T =
+    tryWithContext(s"Error in node type: (${labels.mkString(",")})")(block)
+
+  private def tryWithRel[T](label: String)(block: => T): T =
```
Mats-SX: I like that these were positioned closely.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -467,19 +481,37 @@ case class Join(
...
+}
+
 trait ElementViewKey {
   def elementType: Set[String]
   def qualifiedViewId: QualifiedViewId
 }
 
-case class NodeViewKey(nodeType: Set[String], qualifiedViewId: QualifiedViewId) extends ElementViewKey {
-  override val elementType: Set[String] = nodeType
+case class NodeViewKey(nodeType: NodeType, qualifiedViewId: QualifiedViewId) extends ElementViewKey {
```
Mats-SX: Nice improvement.

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlAcceptanceTest.scala
```diff
@@ -328,6 +328,7 @@ class GraphDdlAcceptanceTest extends BaseTestSuite {
         .withRelationshipType("TYPE_1")
         .withRelationshipPropertyKeys("TYPE_2")("prop" -> CTBoolean.nullable)
         .withSchemaPatterns(SchemaPattern("A", "TYPE_1", "B"))
+        .withSchemaPatterns(SchemaPattern(Set("A", "B"), "TYPE_2", Set("C")))
```
Mats-SX: This seems to indicate that we allow "empty" node and rel types; ie they are part of the graph schema, but there's no data in there. We should make sure that this works also when forming graphs and querying those. I don't think we need to fail when a node or rel type is specified without a mapping, but perhaps supporting that is silly?

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlAstFullExampleTests.scala
```diff
@@ -1,55 +0,0 @@
...
-//    parsed.show()
-  }
-
-  it("parses the Census graph DDL") {
-    val censusUrl = getClass.getResource("/census-graph.ddl")
-    val censusDdlString = Source.fromURL(censusUrl).getLines.mkString(Properties.lineSeparator)
-    val parsed = parse(censusDdlString)
-    // TODO: write expectation :)
-//    parsed.show()
-  }
```
Mats-SX: I suppose we don't need these because they are baked into other tests?

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlParserTest.scala
```diff
@@ -580,6 +531,18 @@ class GraphDdlParserTest extends BaseTestSuite with MockitoSugar with TestNameFi
         parse(ddlString)
       }
     }
+
+    it("gives a nice error message for old syntax") {
```
Mats-SX: Nice! :)

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
```diff
@@ -190,8 +191,132 @@ class GraphDdlTest extends FunSpec with Matchers {
...
 
+  it("allows these equivalent graph definitions") {
+    val ddls = List(
+      // most compact form
+      GraphDdl("""SET SCHEMA ds1.db1
+                 |CREATE GRAPH myGraph (
+                 |  A (x STRING), B (y STRING),
+                 |  (A), (A)-[B]->(A),
+                 |  (A) FROM a,
+                 |  (A)-[B]->(A) FROM b e
```
Mats-SX: This isn't the most compact since it repeats the node and rel types -- or am I missing something here?

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
```diff
@@ -190,8 +191,132 @@ class GraphDdlTest extends FunSpec with Matchers {
...
+                 |    START NODES (A) FROM a n JOIN ON e.id = n.id
+                 |    END   NODES (A) FROM a n JOIN ON e.id = n.id
+                 |)
+               """.stripMargin)
+    )
+
+    ddls(1) shouldEqual ddls.head
+    ddls(2) shouldEqual ddls.head
+    ddls(3) shouldEqual ddls.head
+    ddls(4) shouldEqual ddls.head
```
Mats-SX: Missing case 5.

```suggestion
    ddls(4) shouldEqual ddls.head
    ddls(5) shouldEqual ddls.head
```

or even 

```
    ddls.tail.foreach { ddl =>
      ddl shouldEqual ddls.head
    }
```

(won't bother `suggestion`ing that)

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
```diff
@@ -190,8 +191,132 @@ class GraphDdlTest extends FunSpec with Matchers {
...
+  }
+
+  it("allows these equivalent graph type definitions") {
+    val ddls = List(
+      // most compact form
```
Mats-SX: this isn't the most compact; it's the most compact _that uses a graph type_.

The most compact would be

```
      GraphDdl("""
                 |CREATE GRAPH myGraph (
                 |  A (x STRING), B (y STRING), C (z STRING),
                 |  (A)-[B]->(C) -- no mapping
                 |)
               """.stripMargin),
```

spark-cypher-examples/src/main/resources/northwind/ddl/northwind.ddl
```diff
@@ -212,8 +198,7 @@ CREATE GRAPH Northwind OF NORTHWIND_NAIVE (
   (Supplier)
        FROM VIEW_SUPPLIERS,
 
-  -- (Product)-[HAS_CATEGORY]->(Category)
-  [HAS_CATEGORY]
+  (Product)-[HAS_CATEGORY]->(Category)
```
Mats-SX: I like how we had to add this comment to keep things in mind; and now we made the comment actual syntax :D

spark-cypher-examples/src/main/scala/org/opencypher/spark/util/LdbcUtil.scala
```diff
@@ -52,6 +52,15 @@ object LdbcUtil {
...
     "person_speaks_language")
 
+  case class NodeType(nodeTable: String) {
+    override def toString: String = s"(${nodeTable.toNodeLabel})"
+  }
+
+  case class EdgeType(startNodeType: NodeType, edgeType: String, endNodeType: NodeType) {
+    def edgeTable: String = s"${startNodeType.nodeTable}_${edgeType}_${endNodeType.nodeTable}"
+    override def toString: String = s"$startNodeType-[$edgeType]->$endNodeType"
+  }
```
Mats-SX: Could we perhaps reuse `NodeType` and `RelationshipType` classes from above here? We could have `implicit class` wrappers if we need a different `toString()` shape.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
s1ck: It's following the design for all the conversion from DDL AST nodes to DDL IR nodes. We would need to change it for all the IR classes then.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
s1ck: Same as above. The idea is good. We will create a card and refactor all the transformation methods in a separate PR.

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
s1ck: The suggested method would not tell you which one fails (the line where it fails).

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
s1ck: Yes, but the test is about graph type definitions. We need the empty graph definition to expose the schema.

spark-cypher-examples/src/main/resources/northwind/ddl/northwind.ddl
s1ck: We had the same thought, when we updated the syntax :)

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlAcceptanceTest.scala
s1ck: The CAPS graph is built solely from the node and rel mappings, i.e., this should not be an issue. Having a node/rel type in the schema with no data is basically already possible with having a node/rel type pointing to an empty table. However, it feels a bit inconsistent, not sure how to react on this.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
s1ck: ¯\_(ツ)_/¯

spark-cypher-examples/src/main/scala/org/opencypher/spark/util/LdbcUtil.scala
s1ck: The only interface is the DDL syntax and not the IR. We were thinking about changing the generator to produce DDL IR instead of DDL string. Then it would make sense to reuse the classes.

graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala
Mats-SX: Could get fancy with zipWithIndex but why bother

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
Mats-SX: OK

PR #732, opened 2018-12-05
Death of the SingleTableGraph

DarthMax:
Refactor graph construction planner to minimize necessary `DISTINCT` calls

- construct now creates `ScanGraph`s instead of `SingleTableGraphs`


okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/ConstructGraphPlanner.scala
```diff
@@ -271,4 +275,227 @@ object ConstructGraphPlanner {
...
+      case (variable, _) => variable
+    }.map {
+      case (variable, list) => variable -> list.flatMap(_._2).toSet
+    }
+
+    // Compute Scans from constructed and cloned entities
+    val createdEntityScanVars = scanVarsFromCreatedEntities(logicalPatternGraph)
+    val clonedEntityScanVars = scanVarsFromClonedEntities(logicalPatternGraph)
+
+    val scansForCreated = createScans(createdEntityScanVars, setLabelItems, inputOp, schema)
```
s1ck: Maybe add a short explanation on why can skip the distinct for create items

PR #739, opened 2018-12-13
SQL PGDS file support

DarthMax:
Makes it possible to specify files inside the SQL DDL


spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+  it("reads nodes and rels from file-based sources") {
+    val ddlString =
+      s"""
+         |CREATE GRAPH TYPE personSchema (
+         | Person (id INTEGER, name STRING),
+         | KNOWS,
+         | (Person),
+         | (Person)-[KNOWS]->(Person)
+         |)
+         |CREATE GRAPH fooGraph OF personSchema (
```
Mats-SX: We could simplify by inlining the graph type here.

```suggestion
         |CREATE GRAPH fooGraph (
         | Person (id INTEGER, name STRING),
         | KNOWS,
         |
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
     val e = the [SqlDataSourceConfigException] thrownBy pgds.graph(GraphName("g"))
     e.getMessage should (include("unknown") and include("known1") and include("known2"))
   }
+
+  it("reads nodes and rels from file-based sources") {
+    val ddlString =
+      s"""
+         |CREATE GRAPH TYPE personSchema (
```
Mats-SX:
```suggestion
```

batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         | Person (id INTEGER, name STRING),
```
Mats-SX:
```suggestion
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         | KNOWS,
```
Mats-SX:
```suggestion
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         | (Person),
```
Mats-SX:
```suggestion
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         | (Person)-[KNOWS]->(Person)
```
Mats-SX:
```suggestion
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         |)
```
Mats-SX:
```suggestion
```
batch 1

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         |    FROM csv.`KNOWS.csv` edge
+         |      START NODES (Person) FROM csv.`Person.csv` person JOIN ON person.id = edge.p1
+         |      END   NODES (Person) FROM csv.`Person.csv` person JOIN ON edge.p2 = person.id
+         |)
+     """.stripMargin
+
+    val csvDataSourceConfig = SqlDataSourceConfig(
+      storageFormat = CsvFormat,
+      dataSourceName = "csv",
+      basePath = Some(getClass.getResource("/csv").getPath)
```
Mats-SX: Neat way of using the configuration!

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+         |
+         |  (Person)-[KNOWS]->(Person)
+         |    FROM csv.`KNOWS.csv` edge
+         |      START NODES (Person) FROM csv.`Person.csv` person JOIN ON person.id = edge.p1
+         |      END   NODES (Person) FROM csv.`Person.csv` person JOIN ON edge.p2 = person.id
+         |)
+     """.stripMargin
+
+    val csvDataSourceConfig = SqlDataSourceConfig(
+      storageFormat = CsvFormat,
```
Mats-SX: Is there a reason we chose CSV format only? Should we create mirrors of this test for `Orc` and `Parquet` formats? Possibly we could parameterise them if they end up just diffing on one or two lines.

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -518,4 +518,44 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
+    ds.graph(fooGraphName).nodes("n").toMapsWithCollectedEntities should equal(Bag(
+      CypherMap("n" -> CAPSNode(0, Set("Person"), CypherMap("id" -> 1, "name" -> "Alice"))),
+      CypherMap("n" -> CAPSNode(1, Set("Person"), CypherMap("id" -> 2, "name" -> "Bob"))),
+      CypherMap("n" -> CAPSNode(2, Set("Person"), CypherMap("id" -> 3, "name" -> "Eve")))
+    ))
+
+    ds.graph(fooGraphName).relationships("r").toMapsWithCollectedEntities should equal(Bag(
+      CypherMap("r" -> CAPSRelationship(0, 0, 1, "KNOWS")),
+      CypherMap("r" -> CAPSRelationship(1, 1, 2, "KNOWS"))
+    ))
```
Mats-SX: Maybe it's overkill, but it would be nice to have a Cypher query here as well. Could be something that's easy to assert on, like an aggregation:

```
ds.graph.cypher("MATCH (n:Person)-->(:Person) RETURN max(n.name) AS max").toMapsWithCollectedEntities should equal(Bag(
  CypherMap("max" -> "Eve")
))
```

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlDataSourceConfig.scala
```diff
@@ -58,14 +58,16 @@ object SqlDataSourceConfig {
...
   */
 case class SqlDataSourceConfig(
   storageFormat: StorageFormat,
   dataSourceName: String,
   defaultSchema: Option[String] = None,
   jdbcUri: Option[String] = None,
   jdbcDriver: Option[String] = None,
-  jdbcFetchSize: Int = 100
+  jdbcFetchSize: Int = 100,
+  basePath: Option[String] = None
```
Mats-SX: I'm wondering if we should think a bit about how this will scale with additional future configuration formats. JDBC and `basePath` wouldn't be used in conjunction, so perhaps we ought to move to a hierarchical model instead of a flat with many empty fields? Needn't be done on this PR, but perhaps next config variant we come up with.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -168,18 +170,42 @@ case class SqlPropertyGraphDataSource(
...
-    val tableName = qualifiedViewId.schema + "." + qualifiedViewId.view
+
     val inputTable = sqlDataSourceConfig.storageFormat match {
+      case JdbcFormat | HiveFormat =>
+        readSqlTable(viewId, sqlDataSourceConfig)
+
+      case ParquetFormat | CsvFormat | OrcFormat =>
+        readFile(viewId, sqlDataSourceConfig)
+
+
```
Mats-SX:
```suggestion

```
batch 2

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -168,18 +170,42 @@ case class SqlPropertyGraphDataSource(
...
+  private def readSqlTable(viewId: ViewId, sqlDataSourceConfig: SqlDataSourceConfig) = {
+    val spark = caps.sparkSession
+
+    val tableName = (viewId.maybeSetSchema, viewId.parts) match {
+      case (_, dataSource :: schema :: view :: Nil) => s"$schema.$view"
+      case (Some(SetSchemaDefinition(dataSource, schema)), view :: Nil) => s"$schema.$view"
+      case (None, view) if view.size < 3 =>
+        malformed("Relative view identifier requires a preceding SET SCHEMA statement", view.mkString("."))
+      case (Some(_), view) if view.size > 1 =>
+        malformed("Relative view identifier must have exactly one segment", view.mkString("."))
```
Mats-SX: This [looks like duplication](https://github.com/opencypher/cypher-for-apache-spark/pull/739/files#diff-39a5a37a7c93f2de2944037664e769eeR422) from `ViewId.datasource`; could we not use that here?

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -190,13 +216,40 @@ case class SqlPropertyGraphDataSource(
...
+    val optionsByFormat: Map[StorageFormat, Map[String, String]] = Map(
+      CsvFormat -> Map("header" -> "true", "inferSchema" -> "true"),
+      OrcFormat -> Map.empty,
+      ParquetFormat -> Map.empty
+    )
+
+    val viewPath = (viewId.maybeSetSchema, viewId.parts) match {
+      case (Some(_), path :: Nil) => Path(path)
+      case (None, ds :: path :: Nil) => Path(path)
+      case _ => malformed("File names must be defined with the data source", viewId.parts.mkString("."))
```
Mats-SX: Looks similar to what `ViewId.datasource` does; perhaps we should move this into the class as a `def`?

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -190,13 +216,40 @@ case class SqlPropertyGraphDataSource(
...
+    }
+
+    val filePath = if (viewPath.isAbsolute) {
+      viewPath
```
Mats-SX: I don't think I spotted a test with an absolute path; we should definitely have one.

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
DarthMax: There are acceptance tests for parquet, csv, orc. In general there is no difference between those, it is just using another type string in the `SparkSession.read` call

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
DarthMax: Nope at the point of ViewId.datasource it is not possible any more to determine if the view reference is valid or not as it depends on the datasource config. Sql based source accept `a.b.c` or `a`, while the file based accepts only `a.b`

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
Mats-SX: But where is this parameter changed then? The `storageFormat = CsvFormat`?

DarthMax: It's not in those tests. The SQL PGDS implements the PGDS acceptance tests for various formats (including csv, orc, and parquet)

PR #744, opened 2018-12-18
Make Customer360 a real example with a proper test

Mats-SX:
Working on this discovered a bug with MonotonicallyIncreasingId; not sure exactly what, but could be the thing where we have to order the table on all partitions first.


pstutz:
Bug mitigated in https://github.com/opencypher/cypher-for-apache-spark/pull/749

The code that creates Hive views uses Spark SQL and therefore the buggy Spark distinct implementation. We tried creating tables from DFs instead, which allows us to use our own distinct implementation. Unfortunately that makes the test run very slowly, because many tables need to be stored to disk.

Options are to:
1. Merge with SQL query for view creation, only HashId strategy enabled/working. This has the drawback that switching to MonotonicallyIncreasingId strategy would still be incorrect.
2. Use the stored DFs as the testing basis and have both strategies working. This has the drawback that the tests are so painfully slow that we didn't even await running them like this to verify that both strategies produce the expected result, and we'd pay that same waiting time for every CI run.
3. Do not merge for now and either later rethink the test setup or await the Spark bug fix.

Given the drawbacks of the first two options and a lack of urgency we opted for 3. for now. I will still create a separate issue to verify that the bug mitigation works in the bug investigation [PR](https://github.com/opencypher/cypher-for-apache-spark/pull/747) and extract it into a test that runs faster.

PR #749, opened 2019-01-09
Distinct workaround

soerenreichardt:
Workaround for a bug in spark: https://issues.apache.org/jira/browse/SPARK-26572

Implement distinct in CAPS on SparkDataFrame with groupBy and aggregates to avoid the bug.

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/SparkTableTest.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/SparkTableTest.scala
@@ -26,11 +26,12 @@
  */
 package org.opencypher.spark.impl
 
-import org.apache.spark.sql.Row
+import org.apache.spark.sql.{DataFrame, Row, functions}
 import org.apache.spark.sql.types._
 import org.opencypher.okapi.testing.Bag
 import org.opencypher.okapi.testing.Bag._
 import org.opencypher.spark.impl.DataFrameOps._
+import org.opencypher.spark.impl.table.SparkTable.DataFrameTable
 import org.opencypher.spark.testing.CAPSTestSuite
 import org.scalatest.Matchers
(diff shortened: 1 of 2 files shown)


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/SparkTableTest.scala
```diff
@@ -73,4 +76,40 @@ class DataFrameOpsTest extends CAPSTestSuite with Matchers with GeneratorDrivenP
...
+    }
+
+    it("distinct on subset of columns") {
+      val table: DataFrameTable = Seq((1, 2), (1, 3)).toDF("first", "second")
+      val result = table.distinct("first").df
+      result.count() shouldBe 1
+      result.columns.toSeq should equal(table.df.columns.toSeq)
+    }
+
+    it("works around the spark bug") {
```
tobias-johansson: I don't follow what this test does :)
Are we demonstrating the bug or are we demonstrating the workaround?
Could we add even more comments on what it's doing, expected/erroneous outputs etc?

tobias-johansson: If it is indeed demonstrating the workaround, perhaps we should add a test that would fail whenever the bug gets fixed upstream so we know to remove the workaroud?

pstutz: We built an alternative distinct implementation that doesn't use the actual Spark `distinct` as a workaround for the bug. Here we merely check that the semantics for that implementation match the usual `distinct` semantics of the Spark version.

pstutz: That's a good idea, adding a test that fails when the bug is fixed and putting comments with instructions how to remove the workaround next to it.

Issue #758, opened 2019-01-15
FSPGDS cannot store nodes without labels


pstutz:
Reported by @tobias-johansson:

```
The following seems to break FsPgds at least

CATALOG CREATE GRAPH fs.foo {
  CONSTRUCT
    CREATE (:S)-[:R]->()
  RETURN GRAPH
}
since part of the path is labels.mkString("_") which leads to file name collisions


Another thing: 
The resulting exception is thrown from a Future
 and seems to not get propagated back to the calling thread.
This makes me think that there is something fishy with the
Future handling in AbstractPropertyGraphDataSource
```

PR #759, opened 2019-01-15
Add support for nodes without labels to FSPGDS

pstutz:
Fixes https://github.com/opencypher/cypher-for-apache-spark/issues/758

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
@@ -56,20 +56,20 @@ object DefaultGraphDirectoryStructure {
     def path: String = graphName.value.replace(".", pathSeparator)
   }
 
-  // Because an empty path does not work, we need a special directory name for nodes without labels.
-  val noLabelNodeDirectoryName: String = "__NO_LABEL__"
-
   val pathSeparator: String = Path.SEPARATOR
 
   val schemaFileName: String = "propertyGraphSchema.json"
 
   val capsMetaDataFileName: String = "capsGraphMetaData.json"
 
-  val nodeTablesDirectory = "nodes"
+  val nodeTablesDirectoryName = "nodes"
 
-  val relationshipTablesDirectory = "relationships"
+  val relationshipTablesDirectoryName = "relationships"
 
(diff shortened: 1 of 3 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
```diff
@@ -56,6 +56,9 @@ object DefaultGraphDirectoryStructure {
     def path: String = graphName.value.replace(".", pathSeparator)
   }
 
+  // Because an empty path does not work, we need a special directory name for nodes without labels.
+  val noLabelNodeDirectoryName: String = "NO_LABEL_NODES"
```
tobias-johansson: Does this mean that `()` and `(:NO_LABEL_NODES)` or `(:NO:LABEL:NODES)` would map to the same file?
Looking at `encodeSpecialCharacters` it seems like that would be the case, no?

pstutz: AFAIK the labels are sorted alphabetically, so only the first case should be problematic. I guess we could opt for making the special case even longer and less collision-prone or add an extra marker to all the other directory names. I don't like the first one because it's ugly (and I already chose a long name because of that) and the second one because it would not be backwards-compatible. Do you think that the possible collision with the `:NO_LABEL_NODES` label is something we should mitigate?

pstutz:
Currently blocked because the DDL does not allow defining nodes without labels.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
tobias-johansson: We might do something like `__NO_LABEL__`?
Perhaps the possible collisions here aren't probable enough to warrant spending time on an even better scheme right now :)

If it really becomes a problem it should be possible implement a whole new scheme in parallel, and switch based on some marker.

pstutz: Will switch it to `__NO_LABEL__`

pstutz:
Resolved the DDL issue by adding a custom wrapper on top of the SQLPGDS that adds the unsupported nodes without labels.

PR #760, opened 2019-01-15
Planning optimizations

s1ck:


spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -254,16 +255,17 @@ case class SqlPropertyGraphDataSource(
...
+    val fields = dataFrame.schema.fields
+    val indexedFields = fields.map(field => field.name.toLowerCase).zipWithIndex.toMap
 
-    mapping.propertyMapping.foldLeft(dataFrame) {
-      case (currentDf, (property, column)) if dfColumns.contains(column.toLowerCase) =>
-        currentDf.withColumnRenamed(column, property.toPropertyColumnName)
-      case (_, (_, column)) => throw IllegalArgumentException(
+    val columnRenamings = mapping.propertyMapping.map {
+      case (property, column) if indexedFields.contains(column.toLowerCase) =>
+        fields(indexedFields(column.toLowerCase)).name -> property.toPropertyColumnName
```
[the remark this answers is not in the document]
s1ck: We did, but I was a bit confused that this actually worked. Before that fix, the two JDBC bases examples (Northwind and Census) failed because the column names were upper case and not renamed within that method.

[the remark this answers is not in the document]
s1ck: I opted for tracking the original column (by index) and rename based on that column name.

spark-cypher-testing/src/main/scala/org/opencypher/spark/testing/support/RecordMatchingTestSupport.scala
```diff
@@ -43,6 +48,21 @@ trait RecordMatchingTestSupport {
...
 
+  implicit class RichRow(r: Row) {
+
+    def getCypherValue(expr: Expr, header: RecordHeader)
+      (implicit context: RelationalRuntimeContext[DataFrameTable]): CypherValue = {
+      expr match {
+        case Param(name) => context.parameters(name)
+        case _ =>
+          header.getColumn(expr) match {
+            case None => throw IllegalArgumentException(s"column for $expr")
```
pstutz: Would be helpful to give more context here about the header that didn't have this expression.

spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
```diff
@@ -277,4 +281,248 @@ object SparkTable {
     override def columnsFor(returnItem: String): Set[String] =
       throw UnsupportedOperationException("A DataFrameTable does not have return items")
   }
+
+  implicit class DataFrameMeta(val df: DataFrame) extends AnyVal {
```
pstutz: Common names would be `DataFrameOps`, `RichDataFrame`, `WrappedDataFrame`. We should probably pick a naming pattern for this kind of implicit class and use it consistently.

s1ck: I opted to split up the extension methods into multiple implicit classes, e.g. `DataFrameMeta`, `DataFrameTransformation` etc. `Ops` or `Rich` would be too abstract. Do you have a suggestion?

pstutz: No, I love the more concrete ring of `Meta`. :P

s1ck: :D right ...

PR #765, opened 2019-01-17
Migrate parsers for DDL and Cypher types to FastParse 2.1.0

pstutz:
[FastParse 2.1.0 should be 3-4x faster](https://www.lihaoyi.com/fastparse) and allows us to stay up-to-date cheaply going forward.


pstutz:
Code style is set to `caps-style`, maybe it could be changed to put spaces around the`@` extractors.

PR #767, opened 2019-01-22
Retain semantic direction during IR and logical planning

s1ck:


okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
```diff
@@ -522,11 +524,20 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
...
+        producer.planExpandInto(c.source, r, c.target, Outgoing, sourcePlan)
+
+      case rel: DirectedRelationship if sourcePlan == targetPlan && rel.semanticDirection == INCOMING =>
+        producer.planExpandInto(c.target, r, c.source, Incoming, targetPlan)
+
+      case rel: DirectedRelationship if rel.semanticDirection == OUTGOING =>
+        producer.planExpand(c.source, r, c.target, Outgoing, sourcePlan, targetPlan)
+
+      case rel: DirectedRelationship if rel.semanticDirection == INCOMING =>
+        producer.planExpand(c.target, r, c.source, Incoming, targetPlan, sourcePlan)
```
DarthMax: I suggest keeping  `source` as the rels source and `target` as the rels target, thus only using the DIRECTION to differentiate between incoming and outgouing.

s1ck: Good call. It's cleaner now as the logic is one place!

PR #775, opened 2019-01-25
Enable scalac plugin to fail on `???`

tobias-johansson:
Enables build failure if `???` is encountered in the code

```
$ gradle graph-ddl:classes -Pcfg.ci=true

> Task :graph-ddl:compileScala FAILED
Pruning sources from previous analysis, due to incompatible CompileSetup.
/home/tobias/Projects/cypher-for-apache-spark/graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala:46: Found hole with type: String
  def x: String = ???
                  ^
one error found

FAILURE: Build failed with an exception.
```

Uses https://github.com/cb372/scala-typed-holes, which is at version 0.0.6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,10 @@ subprojects {
         mavenLocal()
     }
 
+    configurations {
+        scalaCompilerPlugin
+    }
+
     dependencies {
         compile group: 'org.scala-lang', name: 'scala-library',  version: ver.scala.full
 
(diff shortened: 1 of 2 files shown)


Mats-SX:
I don't mind that this plugin has such a version. From my POV it is just for convenience, or extra belt+suspenders from review misses (we do pairing and code review to catch most of these things). If there isn't a test that captures a code path we really intend to support we're making a bigger mistake than leaving a `???` in I think.

PR #778, opened 2019-01-29
Remove unnecessary aliasing in df.castToLong helper

### LINKED ISSUE #773: Planner hangs for TPC-H Query 14
```
MATCH
  (l:Lineitem)-[:HAS_PART]->(p:Part)
WHERE
  l.l_shipdate >= 809913600
  AND l.l_shipdate < 812505600
RETURN
  100.0 * SUM(
  CASE
    WHEN p.p_type =~ 'PROMO.*' THEN l.l_extendedprice * (1 - l.l_discount)
    ELSE 0.0
    END
  ) /
  SUM(l.l_extendedprice * (1 - l.l_discount)) AS promo_revenue
```

Relational planner runs through. However, after calling `df.show` or `df.explain`, the program hangs, one thread is running (probably catalyst) and the program crashes eventually (no exception).

A workaround is an additional `WITH` clause.

```
MATCH
  (l:Lineitem)-[:HAS_PART]->(p:Part)
WHERE
  l.l_shipdate >= 809913600
  AND l.l_shipdate < 812505600
WITH
  SUM(
  CASE
    WHEN p.p_type =~ 'PROMO.*' THEN l.l_extendedprice * (1 - l.l_discount)
    ELSE 0.0
    END
  ) /
  SUM(l.l_extendedprice * (1 - l.l_discount)) AS promo_revenue
RETURN promo_revenue * 100.00
```

s1ck:
For TPC-H Query 14 the additional aliasing resulted in an issue within the `InferFiltersFromConstraints` rewriter in Catalyst. Will try to reproduce the bug with plain dataframes and report to Spark Jira.

* fixes #773

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
@@ -422,14 +422,15 @@ object SparkTable {
       */
     def castToLong: DataFrame = {
       def convertColumns(field: StructField, col: Column): Column = {
-        field.dataType match {
+        val convertedCol = field.dataType match {
           case StructType(inner) =>
-            val fields = inner.map(i => convertColumns(i, col.getField(i.name)))
-            functions.struct(fields: _*)
+            val columns = inner.map(i => convertColumns(i, col.getField(i.name)).as(i.name))
+            functions.struct(columns: _*)
           case ArrayType(IntegerType, nullable) => col.cast(ArrayType(LongType, nullable))
           case IntegerType => col.cast(LongType)
           case _ => col
         }
+        if (col == convertedCol) col else convertedCol.as(field.name)
       }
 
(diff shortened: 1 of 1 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
```diff
@@ -373,7 +373,7 @@ object SparkTable {
     }
 
     def safeRenameColumns(renamings: Map[String, String]): DataFrame = {
-      if (renamings.forall { case (oldColumn, newColumn) => oldColumn == newColumn }) {
+      if (renamings.isEmpty || renamings.forall { case (oldColumn, newColumn) => oldColumn == newColumn }) {
```
pstutz: Can't `renamings.isEmpty || ' be removed without changing the semantic?

s1ck: You're right. `Seq.forall` already returns `true` if the `Seq` is empty.

PR #793, opened 2019-02-09
Add Gradle task to run custom Scala app

s1ck:
Allows you to do:
`./gradlew spark-cypher-examples:runApp -PmainClass=org.opencypher.spark.examples.CaseClassExample`

@tobias-johansson maybe you know a better way. I tried the `application` plugin but that also requires passing a system property.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,11 +83,16 @@ subprojects {
         group 'help'
     }
 
-    task runApp(type:JavaExec) {
+    task runApp {
+        dependsOn tasks.compileScala
         group 'run'
         description 'Run a custom Scala app (use -PmainClass=path.to.App)'
-        main = project.hasProperty("mainClass") ? project.getProperty("mainClass") : ""
-        classpath = sourceSets.main.runtimeClasspath
+        doLast {
+            project.javaexec {
+                classpath = sourceSets.main.runtimeClasspath
+                main = project.getProperty("mainClass")
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -82,6 +82,13 @@ subprojects {
         description 'Searches all projects for a dependency'
         group 'help'
     }
+
+    task runApp(type:JavaExec) {
+        group 'run'
+        description 'Run a custom Scala app (use -PmainClass=path.to.App)'
+        main = project.hasProperty("mainClass") ? project.getProperty("mainClass") : ""
```
tobias-johansson: How does this behave if no `mainClass` is given?

tobias-johansson:
Perhaps we should add a note on this in the readme?

build.gradle
s1ck: The JVM fails to run. The problem is, that any other gradle tasks fails if `main` is not being assigned proper. I was thinking about adding a default class like `CaseClassExample` but that requires the task run within a specific module.

tobias-johansson: To suspend the property lookup until execution time (instead of configuration time) we could do something like this:
```groovy
task runApp {
  group 'run'
  description 'Run a custom Scala app (use -PmainClass=path.to.App)'
  doLast {
    javaExec {
      main = project.getProperty("mainClass")
      classpath = sourceSets.main.runtimeClasspath
    }
  }
}
```

PR #794, opened 2019-02-12
Remove checked-in license lists

Mats-SX:
We still validate licenses towards the whitelist, but we do not
need the checked-in files.


Mats-SX:
Should put the generated files into the `META-INF` directory

PR #798, opened 2019-02-12
Towards a more readable TOC

jjaderberg:
If we could take advantage of title abbreviations[1], we could use a long-form title in the running narrative and an abbreviated title in the TOC. Our docs tooling doesn't currently support this, unfortunately.

[1]: https://tdg.docbook.org/tdg/5.1/titleabbrev.html

### DIFF
--- a/documentation/asciidoc/architecture-and-usage.adoc
+++ b/documentation/asciidoc/architecture-and-usage.adoc
@@ -1,5 +1,5 @@
 [[technology]]
-= Technology choices and future directions
+= Technology choices
 
 [abstract]
 --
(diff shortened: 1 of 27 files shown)


s1ck:
We viewed the docs in the browser and realized that the TOC doesn't have the necessary depth to show the new titles. They are displayed on the top left/right corners for navigation purpose. Is it possible to show them in the TOC, too? This would increase usability for the docs pages.

jjaderberg:
@s1ck Thanks! The TOC depth requires a downstream change; but yes the idea is that the updated, shorter headings will show up in the TOC.

PR #804, opened 2019-02-14
Report error message when property has unsupported Cypher type

Mats-SX:
This is first found by the :C node in the test because it lacks the
property which is typed ANY

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/ExpressionTests.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/ExpressionTests.scala
@@ -29,6 +29,7 @@ package org.opencypher.spark.impl.acceptance
 import org.junit.runner.RunWith
 import org.opencypher.okapi.api.value.CypherValue
 import org.opencypher.okapi.api.value.CypherValue.CypherMap
+import org.opencypher.okapi.impl.exception.NotImplementedException
 import org.opencypher.okapi.testing.Bag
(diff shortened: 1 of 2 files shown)


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/ExpressionTests.scala
```diff
@@ -265,6 +266,44 @@ class ExpressionTests extends CAPSTestSuite with ScanGraphInit {
...
+
+    it("reports error message when property has ANY type") {
+      // Given
+      val given = initGraph("""CREATE (:A {val: 'foo'}), (:B {val: 1}), (:C)""")
+
+      // When
+      val result = given.cypher("MATCH (a) RETURN a.val")
+
+      // Then
+      a [NotImplementedException] shouldBe thrownBy(result.records.show)
```
s1ck: I would like if we over time adapt to a bit more fine grained testing on exception messages. That requires us to think a bit more about the actual message from a user perspective and we don't accidently check for the same exception thrown by some other code path. Here is a good example: https://github.com/opencypher/cypher-for-apache-spark/blob/master/graph-ddl/src/test/scala/org/opencypher/graphddl/GraphDdlTest.scala#L607-L619

Mats-SX: I agree. Thanks for the pointer; I knew I had seen a better way of doing this somewhere.

PR #805, opened 2019-02-14
Replace monotonically increasing ID with serialized ID

pstutz:


spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/IdGenerationStrategy.scala
```diff
@@ -29,5 +29,5 @@ package org.opencypher.spark.api.io.sql
 object IdGenerationStrategy extends Enumeration {
   type IdGenerationStrategy = Value
 
-  val MonotonicallyIncreasingId, HashBasedId = Value
+  val SerializedId, HashBasedId = Value
```
tobias-johansson: Should we rename `HashBasedId` to simply `HashedId`?

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -317,36 +263,51 @@ case class SqlPropertyGraphDataSource(
...
-    newIdColumn: String
+    newIdColumn: String,
+    strategy: IdGenerationStrategy,
+    schema: Schema
   ): DataFrame = {
-    val viewLiteral = functions.lit(elementViewKey.viewId.parts.mkString("."))
-    val elementTypeLiterals = elementViewKey.elementType.toSeq.sorted.map(functions.lit)
     val idColumns = idColumnNames.map(dataFrame.col)
-    dataFrame.withHashColumn(Seq(viewLiteral) ++ elementTypeLiterals ++ idColumns, newIdColumn)
+    strategy match {
```
tobias-johansson: Did you consider moving these cases out into methods on the `IdGenerationStrategy`?
Could be made possible to supply custom strategies that way (or parameterize the hash stategy etc.), but I'm not sure it makes sense to open that up

s1ck: We did consider it, but weren't sure if this is necessary atm.

PR #816, opened 2019-02-19
Replace Neo4jGraphFactory with CreateGraphFactory

s1ck:
Also removes Neo4j dependencies from `okapi-testing`


s1ck:
`Scenario "Should add or subtract duration to or from date #3"` uses floating point numbers as duration values. This worked before, when we used Neo4j for `InitQuery` since Neo4j [converts such values](https://github.com/neo4j/neo4j/blob/3.5/community/values/src/main/java/org/neo4j/values/storable/DurationValue.java#L162-L172). CAPS does not support this yet.

PR #817, opened 2019-02-21
Use specialized types for CypherRecords and CypherResult

s1ck:
Co-authored-by: Philip Stutz <philip.stutz@gmail.com>

### DIFF
--- a/spark-cypher-testing/build.gradle
+++ b/spark-cypher-testing/build.gradle
@@ -8,7 +8,6 @@ dependencies {
 
     compile group: 'org.apache.spark',     name: "spark-core".scala(),     version: ver.spark
     compile group: 'org.apache.spark',     name: "spark-sql".scala(),      version: ver.spark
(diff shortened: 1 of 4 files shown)


spark-cypher-testing/build.gradle
```diff
@@ -8,7 +8,6 @@ dependencies {
 
     compile group: 'org.apache.spark',     name: "spark-core".scala(),     version: ver.spark
     compile group: 'org.apache.spark',     name: "spark-sql".scala(),      version: ver.spark
-    compile group: 'org.apache.spark',     name: "spark-catalyst".scala(), version: ver.spark
```
Mats-SX: Oh really? I guess this doesn't have any runtime effect since Catalyst will (likely) be present in the actual runtime environment. I'm surprised though.

s1ck: It comes with `spark-sql`, but we were surprised, too.

PR #820, opened 2019-02-21
Implement canonical physical record representation for Cypher results

s1ck:
* by default, CypherResult.records will be aligned with return items
* the previous semantics are preserved by CypherRecords.internalRecords

Simple expressions:
```
caps.cypher("RETURN 1 AS foo, 2 AS bar").records.asCaps.df.columns.toSet should equal(Set("foo", "bar"))
caps.cypher("RETURN true AND false").records.asCaps.df.columns.toSet should equal(Set("true AND false"))
```
Entities:
```
val result = given.cypher("FROM GRAPH foo MATCH (n) RETURN n")
result.records.asCaps.df.columns.toSet should equal(Set("n", "n:L", "n.val"))
```

Co-authored-by: Philip Stutz <philip.stutz@gmail.com>


okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/operators/RelationalOperator.scala
```diff
@@ -330,10 +330,33 @@ final case class Select[T <: Table[T] : TypeTag](
...
-  override lazy val returnItems: Option[Seq[Var]] =
+  override lazy val maybeReturnItems: Option[Seq[Var]] =
     Some(returnExpressions.flatMap(_.owner).collect { case e: Var => e }.distinct)
 }
 
+/**
+  * Renames physical columns to given header expression names.
+  * Ensures that there is a physical column for each return item, i.e. aliases lead to duplicate physical columns.
+  */
+final case class AlignColumnsWithReturnItems[T <: Table[T] : TypeTag](
```
DarthMax: I would suggest to either find a new name for this operator or converting this into a planner using more general relational operators

spark-cypher-examples/src/test/resources/example_outputs/DataFrameOutputExample.out
```diff
@@ -1,7 +1,7 @@
-+----------------+-----------------+
-|_name __ STRING?|b_name __ STRING?|
-+----------------+-----------------+
-|           Alice|              Bob|
-|             Bob|            Carol|
-+----------------+-----------------+
++------+------+
```
DarthMax: Nice!

spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
```diff
@@ -64,11 +64,13 @@ object SparkTable {
...
-    override def select(cols: String*): DataFrameTable = {
-      if (df.columns.toSeq == cols) {
+    override def select(col: (String, String), cols: (String, String)*): DataFrameTable = {
+      val columns = col +: cols
+      if (df.columns.toSeq == columns.map { case (_, alias) => alias }) {
         df
       } else {
-        df.select(cols.map(df.col): _*)
+        // Spark interprets dots in column names as struct accessors. Hence, we need to escape column names by default.
+        df.select(columns.map{ case (colName, alias) => df.col(s"`$colName`").as(alias) }: _*)
```
DarthMax: maybe use column sequences instead of string sequences here in order to make selecting nested columns more explicit

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/operators/RelationalOperator.scala
s1ck: Will be addressed in https://github.com/opencypher/cypher-for-apache-spark/issues/821

PR #824, opened 2019-02-25
Return canonical column names for PG.nodes/PG.relationships

pstutz:


pstutz:
Too many PGDS' do not like dots in column names for this approach to be viable.

PR #826, opened 2019-02-26
Produce correct tables in SQL PGDS acceptance tests

s1ck:
The previous logic created a table for each reltype (e.g. `R`) according
to the OKAPI schema. However, `(A)-[R]->(B) `and `(A)-[R]->(C)` are
different rel types in DDL/SQL PGDS world. We therefore need to create
tables for each DDL rel Type in order to create non-overlapping input
tables. This is achieved by joining relation tables with their incident
node tables to get the actual relationships.

### DIFF
--- a/spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/FullPGDSAcceptanceTest.scala
+++ b/spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/FullPGDSAcceptanceTest.scala
@@ -62,11 +62,7 @@ class FullPGDSAcceptanceTest extends CAPSTestSuite
 
   executeScenariosWithContext(allScenarios, SessionContextFactory)
 
-  val sqlBlackList: Set[String] = Set.empty
-
-  allSqlContextFactories.foreach(ctx =>
-    executeScenariosWithContext(allScenarios.filter(s => !sqlBlackList.contains(s.name)), ctx)
-  )
+  allSqlContextFactories.foreach(executeScenariosWithContext(allScenarios, _))
 
   allFileSystemContextFactories.foreach(executeScenariosWithContext(allScenarios, _))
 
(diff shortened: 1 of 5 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/api/io/fs/GraphDirectoryStructure.scala
```diff
@@ -73,12 +73,13 @@ object DefaultGraphDirectoryStructure {
...
       noLabelNodeDirectoryName
     } else {
-      labels.toSeq.sorted.mkString("_").encodeSpecialCharacters
+      concatDirectoryNames(labels.toSeq.sorted)
     }
   }
 
   def relKeyTableDirectoryName(relKey: String): String = relKey.encodeSpecialCharacters
 
+  def concatDirectoryNames(seq: Seq[String]): String = seq.mkString("_").encodeSpecialCharacters
```
pstutz: Does this do the right thing for the no-label case?

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/FullPGDSAcceptanceTest.scala
```diff
@@ -63,7 +62,7 @@ class FullPGDSAcceptanceTest extends CAPSTestSuite
 
   executeScenariosWithContext(allScenarios, SessionContextFactory)
 
-  val sqlBlackList: Set[String] = Set("API: PropertyGraphDataSource: correct node/rel count for graph #1")
+  val sqlBlackList: Set[String] = Set.empty
```
pstutz: We can probably just remove the blacklist now.

s1ck:
> Thanks a lot for this fix.
> 
> So there was never an issue with the DDL or SQL PGDS, but only with how we generated the tables for the initial test graphs?

Yes. We can add validation in the GraphDDL IR construction that checks if the same view is not used for multiple rel types (same for nodes).

PR #833, opened 2019-02-27
Improve type inference for functions

tobias-johansson:
Especially for null and nullability


okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/fromFrontendType.scala
```diff
@@ -72,5 +76,72 @@ object SignatureConverter {
         Some(FunctionSignature(sigInputTypes, sigOutputType))
       }
     }
+
+    def convertDirect: Option[FunctionSignature] = {
```
Mats-SX: This seems very similar to `convert()` above; do we still need both versions?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/fromFrontendType.scala
```diff
@@ -72,5 +76,72 @@ object SignatureConverter {
...
+
+  object FunctionSignatures {
+    def from(original: Seq[TypeSignature]): FunctionSignatures =
+      FunctionSignatures(original.flatMap(_.convertDirect))
+
+    private def mask(size: Int, hits: Int) =
+      Seq.fill(hits)(true) ++ Seq.fill(size - hits)(false)
+
+    private def masks(size: Int, minHits: Int, maxHits: Int) = for {
+      hits <- Range.inclusive(minHits, maxHits)
```
Mats-SX: Maybe call this `hit` as it is just one number? Or is it a bitmask?

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -40,417 +40,403 @@ import org.scalatest.mockito.MockitoSugar
...
-
-    assertExpr.from("date()") shouldHaveInferredType CTDate.nullable
-    assertExpr.from("date('2010-12-10')") shouldHaveInferredType CTDate.nullable
-    assertExpr.from("date({ year: 2018, month: 12, day: 18 })") shouldHaveInferredType CTDate.nullable
+    assertExpr.from("date()") shouldHaveInferredType CTDate
+    assertExpr.from("date(STRING)") shouldHaveInferredType CTDate
+    assertExpr.from("date({ year: 2018, month: 12, day: 18 })") shouldHaveInferredType CTDate
+    assertExpr.from("date(NULL)") shouldHaveInferredType CTNull
+    assertExpr.from("date(STRING_OR_NULL)") shouldHaveInferredType CTDate.nullable
+    assertExpr.from("date(INTEGER_OR_NULL)") shouldHaveInferredType CTNull
```
Mats-SX: Not sure this is a supported signature for `date()`? Should probably fail.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -40,417 +40,403 @@ import org.scalatest.mockito.MockitoSugar
...
 
   it("should type Date") {
-    implicit val context: TypeTracker = typeTracker("d" -> CTDate)
-
-    assertExpr.from("date()") shouldHaveInferredType CTDate.nullable
-    assertExpr.from("date('2010-12-10')") shouldHaveInferredType CTDate.nullable
-    assertExpr.from("date({ year: 2018, month: 12, day: 18 })") shouldHaveInferredType CTDate.nullable
+    assertExpr.from("date()") shouldHaveInferredType CTDate
+    assertExpr.from("date(STRING)") shouldHaveInferredType CTDate
+    assertExpr.from("date({ year: 2018, month: 12, day: 18 })") shouldHaveInferredType CTDate
```
Mats-SX: Should it be like this?
```suggestion
    assertExpr.from("date(MAP)") shouldHaveInferredType CTDate
```

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -40,417 +40,403 @@ import org.scalatest.mockito.MockitoSugar
...
-        | second: 11,
-        | millisecond: 999,
-        | mircosecond: 999,
-        | nanosecond: 999})""".stripMargin) shouldHaveInferredType CTLocalDateTime.nullable
+    assertExpr.from("localdatetime()") shouldHaveInferredType CTLocalDateTime
+    assertExpr.from("localdatetime(STRING)") shouldHaveInferredType CTLocalDateTime
+    assertExpr.from("localdatetime(MAP)") shouldHaveInferredType CTLocalDateTime
+    assertExpr.from("localdatetime(NULL)") shouldHaveInferredType CTNull
+    assertExpr.from("localdatetime(STRING_OR_NULL)") shouldHaveInferredType CTLocalDateTime.nullable
+    assertExpr.from("localdatetime(INTEGER_OR_NULL)") shouldHaveInferredType CTNull
```
Mats-SX: Same comment as for `date()` above. I don't think this is a supported signature.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -490,126 +472,122 @@ class SchemaTyperTest extends BaseTestSuite with Neo4jAstTestSupport with Mockit
...
-        "intParam" -> CypherValue(42L)
-      ))
+    //implicit val context: TypeTracker = TypeTracker.empty
+    //  .updated(Variable("map")(pos), CTMap(Map("foo" -> CTInteger, "bar" -> CTBoolean)))
+    //  .updated(Variable("stringKey")(pos), CTString)
+    //  .updated(Variable("intKey")(pos), CTInteger)
+    //  .withParameters(Map(
+    //    "stringParam" -> CypherValue("bar"),
+    //    "intParam" -> CypherValue(42L)
+    //  ))
```
Mats-SX: Can remove comment

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -490,126 +472,122 @@ class SchemaTyperTest extends BaseTestSuite with Neo4jAstTestSupport with Mockit
...
 
-    assertExpr.from("""map["foo"]""") shouldHaveInferredType CTInteger
-    assertExpr.from("""map["bar"]""") shouldHaveInferredType CTBoolean
-    assertExpr.from("""map["baz"]""") shouldHaveInferredType CTVoid
+    assertExpr.from("""MAP["INTEGER"]""") shouldHaveInferredType CTInteger
+    assertExpr.from("""MAP["BOOLEAN"]""") shouldHaveInferredType CTBoolean
+    assertExpr.from("""MAP["baz"]""") shouldHaveInferredType CTVoid
```
Mats-SX: It would probably be a little easier to read if we used normal double-quotes on the outside (for Scala) and single-quotes (apostrophes) on the inside (for Cypher):

```suggestion
    assertExpr.from("MAP['baz']") shouldHaveInferredType CTVoid
```

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
```diff
@@ -490,126 +472,122 @@ class SchemaTyperTest extends BaseTestSuite with Neo4jAstTestSupport with Mockit
...
-    assertExpr.from("string ENDS WITH r") shouldHaveInferredType CTBoolean.nullable
-    assertExpr.from("string CONTAINS r") shouldHaveInferredType CTBoolean.nullable
+    assertExpr.from("NULL STARTS WITH STRING") shouldHaveInferredType CTNull
+    assertExpr.from("NULL ENDS WITH STRING") shouldHaveInferredType CTNull
+    assertExpr.from("NULL CONTAINS STRING") shouldHaveInferredType CTNull
   }
+}
 
-  private def typeTracker(typings: (String, CypherType)*): TypeTracker =
+protected abstract class SchemaTyperTestSuite extends BaseTestSuite with Neo4jAstTestSupport {
```
Mats-SX: I like how this test class was improved!

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/AggregationTests.scala
```diff
@@ -500,7 +500,7 @@ class AggregationTests extends CAPSTestSuite with ScanGraphInit {
       ))
     }
 
-    ignore("simple sum(prop) with only null values in RETURN") {
+    it("simple sum(prop) with only null values in RETURN") {
```
Mats-SX: Nice wins!

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/NullTests.scala
```diff
@@ -0,0 +1,123 @@
...
+    it("calling: min(null)")(returnsNull())
+    it("calling: sum(null)")(returnsNull())
+  }
+
+  describe("null input produces value") {
+    it("calling: null IS NULL")(returnsValue(true))
+    it("calling: null IS NOT NULL")(returnsValue(false))
+    it("calling: count(null)")(returnsValue(0))
+    it("calling: collect(null)")(returnsValue(List()))
+  }
```
Mats-SX: Very comprehensive!

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -89,6 +89,17 @@ object SparkSQLExprMapper {
...
 
+        case AliasExpr(innerExpr, _) => innerExpr.asSparkSQLExpr
+
+        case Explode(list) => list.cypherType match {
+          case CTList(_) | CTListOrNull(_) => functions.explode(list.asSparkSQLExpr)
+          case CTNull => functions.explode(functions.lit(null).cast(ArrayType(NullType)))
+          case other => throw IllegalArgumentException("CTList", other)
+        }
+
+        // degenerate cases
```
Mats-SX: What's a 'degenerate case'?

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -89,6 +89,17 @@ object SparkSQLExprMapper {
 
       expr match {
 
+        case AliasExpr(innerExpr, _) => innerExpr.asSparkSQLExpr
+
+        case Explode(list) => list.cypherType match {
```
Mats-SX: So we allow aliasing and UNWIND to precede the general `null in null out` rule here?

spark-cypher/src/main/scala/org/opencypher/spark/impl/table/SparkTable.scala
```diff
@@ -189,6 +171,30 @@ object SparkTable {
...
+                .as(columnName)
+
+            case Avg(expr) =>
+              withInnerExpr(expr)(
+                functions
+                  .avg(_)
+                  .cast(cypherType.getSparkType)
+                  .as(columnName))
+
+            case CountStar(_) =>
```
Mats-SX: Could move this up for clarity perhaps? It won't ever be caught by the general `null in null out` case right?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -125,16 +128,21 @@ object SchemaTyper {
     case HasLabels(node, labels) =>
       for {
         nodeType <- process[R](node)
-        result <- nodeType.material match {
+        result<- nodeType.material match {
+          case CTVoid =>
+            updateTyping[R](node -> CTNull) >>
```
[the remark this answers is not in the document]
tobias-johansson: You are right :) We just copied the behavior..

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -175,12 +183,14 @@ object SchemaTyper {
       for {
         lhsType <- process[R](lhs)
         rhsType <- process[R](rhs)
-        result <- rhsType match {
-          case _: CTList | _: CTListOrNull | CTNull =>
-            recordTypes(lhs -> lhsType, rhs -> rhsType) >> recordAndUpdate(expr -> CTBoolean)
-          case x =>
-            error(InvalidType(rhs, CTList(CTWildcard), x))
+        result <- (rhsType, lhsType) match {
```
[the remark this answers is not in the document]
tobias-johansson: :+1:

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -175,12 +183,14 @@ object SchemaTyper {
...
+          case (CTNull, _) => pure[R, CypherType](CTNull)
+          case (_, CTNull) => pure[R, CypherType](CTNull)
```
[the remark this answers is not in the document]
tobias-johansson: Nice catch, I'll add a test and fix

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
```diff
@@ -126,15 +129,20 @@ object SchemaTyper {
...
+          case CTVoid =>
+            updateTyping[R](node -> CTNull) >>
+            pure[R, CypherType](CTNull)
+
           case CTNode(nodeLabels, qgn) =>
             val detailed = nodeLabels ++ labels.map(_.name).toSet
-            recordType[R](node -> nodeType) >>
-              updateTyping[R](node -> CTNode(detailed, qgn)) >>
-              recordAndUpdate[R](expr -> CTBoolean)
+            updateTyping[R](node -> CTNode(detailed, qgn)) >>
```
tobias-johansson: @Mats-SX with regard to your comment above, could you explain the logic behind this line? (note that we did the same here before)
Seems like we're saying that whenever we see a predicate `a:A` we commit that we now know that `a` has label `A` for use in further processing? Depending on how we thread the state (type tracker) around it seems to me that we are either implementing some form of flow-based typing* , or that this is a bug

*- in parts of the program where the predicate is true the type can be known to include label `A`

s1ck: I think this is a bug. `a:A` is a predicate and we cannot guarantee that this is always `true` for `a` nodes.

Mats-SX: I think it is meant as that flow-based concept (although that term is new to me). The only way to find out what labels are on the nodes is to inspect these predicates. It is intended to only affect the subsequent references of that variable, enabling them to properly compute property types of those nodes.

Essentially:

```
MATCH (a)
WITH a.prop // may be difficult to find exact type
WHERE a:Foo
WITH a.prop // we have more information and can do a better schema lookup
...
```

Mats-SX: > I think this is a bug. `a:A` is a predicate and we cannot guarantee that this is always `true` for `a` nodes.

That's true, but no nodes will be bound to `a` where the predicate does not hold.

s1ck: Actually, it's ok if `n:A` is the only predicate (the resulting `CTNode(:A)` is used for the scan etc. It's different when it's `n:A OR n:B`. The case for `Ors` in the `SchemaTyper` handles `HasLabels` differently, i.e. removes the label from the type and plans a separate filter instead.

tobias-johansson: I also think it's a bug.
Consider the case `MATCH (a) RETURN a:A, a` or more generally `MATCH (a) RETURN a:A, e` where `e` is some expression that references `a`. In this case we shouldn't assert inside `e` that `a` has label `A`, right?
So the question is then would type inference of `e` see the updated context or the previous one?

s1ck: I was just looking (again) into `MATCH (n) WHERE n:A or n:B RETURN n` where the "knownTypes" has precedence. So it actually derives `n -> CTNode(:A:B)` when going through the predicates. But since the pattern says, `n` is just `CTNode` this doesn't matter.

s1ck: This scenario is a bit weird:
```
val given = initGraph("CREATE (:A),(:B)")
val result = given.cypher("MATCH (n:A) WHERE n:A or n:B RETURN n")
result.show
```
Returns only the `:A` node. Seems like predicate within the pattern has precedence.

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
soerenreichardt: Yes that's the reason

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/fromFrontendType.scala
tobias-johansson: `hits` contains the current number of hits, so I think it should be plural

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyperTest.scala
tobias-johansson: Neo4j evaluates `RETURN date([1, null][1])` successfully to `null`
By the premise that the typesystem should only rule out programs that are proven to never work, this should be allowed, right?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/parse/functions/FunctionExtensions.scala
```diff
@@ -78,4 +84,45 @@ case object Duration extends Function with TypeSignatures {
...
+  override def toNeoTypeString: String = "IDENTITY"
+}
+
+case object Id extends Function with TypeSignatures {
+  def name: String = functions.Id.name
+
+  override val signatures: Vector[TypeSignature] = Vector(
+    TypeSignature(argumentTypes = Vector(CTNode), outputType = CTIdentity),
+    TypeSignature(argumentTypes = Vector(CTRelationship), outputType = CTIdentity)
+  )
```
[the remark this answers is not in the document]
tobias-johansson: Created a card

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
tobias-johansson: An edge case where the normal procedure is not required or possible, or something along those lines :)

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/typer/SchemaTyper.scala
tobias-johansson: Let's revisit this at a later point :)

PR #842, opened 2019-02-28
Fix property-based tests

pstutz:


spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/values/CAPSLiteralTests.scala
```diff
@@ -40,49 +39,35 @@ import org.scalatest.prop.Checkers
 class CAPSLiteralTests extends CAPSTestSuite with Checkers with ScanGraphInit {
```
tobias-johansson: Would it be possible to use the scalatest `PropertyChecks` here and get the simpler syntax
```
it("something") {
  forAll(supportedLiteral) { v => ... }
}
```
or does that not play well with `claimant`?

pstutz: I think that does work and would be compatible with `claimant`. `check` allows you to specify `minSuccessful`, not sure if/how one can configure it without it.

tobias-johansson: Looks like config can be passed like this:
```scala
forAll(Gen.identifier, minSuccessful(100)) { id =>
  id.length shouldBe 0
}
```

But I don't know how to pass in `claimant` `Prop`s
For instance, this passes on all examples:
```scala
forAll(Gen.identifier, minSuccessful(100)) { id =>
  Claim(id.length == 0)
}
```

pstutz: Thanks for verifying this. Should we maybe get rid of `claimant`?

PR #843, opened 2019-02-28
Remove Ternary and CTWildcard

s1ck:
Co-authored-by: Max Kießling <max.kiessling@neotechnology.com>


okapi-api/src/test/scala/org/opencypher/okapi/api/types/CypherTypesTest.scala
```diff
@@ -303,51 +275,23 @@ class CypherTypesTest extends ApiBaseTest {
   }
 
   it("type equality between different types") {
```
soerenreichardt: The "between" should be changed to "for" to make clear that we are only comparing the same types with each other.

DarthMax:
We kept the default only for the very common Var case that is used in too many tests. For every other case we deliberately opted for removing those defaults to make it more visible that the correct type needs to be supplied

PR #844, opened 2019-03-01
Change qualifiers and task names for fat jars

tobias-johansson:


spark-cypher/build.gradle
```diff
@@ -14,8 +14,11 @@ dependencies {
     compileOnly group: 'org.apache.spark', name: "spark-catalyst".scala(), version: ver.spark
 }
 
-shadowJar {
-    classifier = 'cluster'
+task allJar(type: ShadowJar) {
+    classifier = 'all'
+
+    from project.sourceSets.main.output
+    configurations = [project.configurations.runtime]
```
Mats-SX: The two tasks look very similar; is it possible to reuse the configuration somehow?

tobias-johansson: Yeah, not in any simple way really

PR #849, opened 2019-03-04
Return unit records and plan filter for strict MATCH

s1ck:
Co-authored-by: Tobias Johansson <tobias.johansson@neotechnology.com>


okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalOperator.scala
```diff
@@ -223,7 +223,13 @@ final case class ExpandInto(
   extends StackingLogicalOperator
     with ExpandOperator {
 
-  override val fields: Set[Var] = lhs.fields ++ rhs.fields
+  override val fields: Set[Var] = {
+    val lhsRhs = lhs.fields ++ rhs.fields
+    rel match {
```
DarthMax: No match is needed here, rel is already a `Var`

s1ck: Thanks, we just copy pasted it from `Expand` without further thinking. Changed in both places.

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
```diff
@@ -128,7 +129,15 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
...
         val patternPlan = planMatchPattern(inputGraphPlan, pattern, where, graph)
-        if (optional) producer.planOptional(inputGraphPlan, patternPlan) else patternPlan
+        if (optional) {
+          producer.planOptional(inputGraphPlan, patternPlan)
+        } else {
+          // We need to filter out unit records for strict MATCH
+          import Expr._
+          val fields = patternPlan.fields intersect pattern.fields.map(toVar)
+          val fieldsNotNullExprs = fields.toSeq.sorted.map(f => IsNotNull(f)(CTBoolean))
+          planFilter(patternPlan, ListSet(fieldsNotNullExprs: _*))
```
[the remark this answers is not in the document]
Mats-SX: I mean, when we plan a node scan we take another plan as input; node scans are not leaf plans. That input plan produces a binding table that is either the unit table (first MATCH) or the result of a previous MATCH. These are either joined together (if the patterns are connected, which is never the case for the first MATCH) or yield a cartesian product. Wouldn't this cartesian product make the unit row disappear automatically?

I thought cartesian worked like this:
- for each row in lhs, take all rows in rhs and append to this row
There is only 1 row in lhs, and 0 columns => 
    (lhs CARTESIAN rhs) := rhs iff lhs == unit table

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
```diff
@@ -45,14 +46,11 @@ case class CAPSRecordsFactory(implicit caps: CAPSSession) extends RelationalCyph
...
 
-  override def unit(): CAPSRecords = {
-    val initialDataFrame = caps.sparkSession.createDataFrame(Seq(EmptyRow()))
-    CAPSRecords(RecordHeader.empty, initialDataFrame)
-  }
-
-  override def empty(initialHeader: RecordHeader = RecordHeader.empty): CAPSRecords = {
-    val initialSparkStructType = initialHeader.toStructType
-    val initialDataFrame = caps.sparkSession.createDataFrame(Collections.emptyList[Row](), initialSparkStructType)
+  override def unit(initialHeader: RecordHeader = RecordHeader.empty): CAPSRecords = {
```
Mats-SX: Why does this take a header as argument? It's not a unit table if it has columns.

spark-cypher-testing/src/test/scala/org/opencypher/spark/SparkTests.scala
```diff
@@ -32,10 +32,6 @@ import org.opencypher.spark.testing.CAPSTestSuite
 
 class SparkTests extends CAPSTestSuite {
 
-  it("foo") {
```
Mats-SX: Oops, that looks like me!

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/operators/RelationalOperator.scala
```diff
@@ -406,7 +405,7 @@ final case class EmptyRecords[T <: Table[T] : TypeTag](
 
   override lazy val header: RecordHeader = RecordHeader.from(fields)
 
-  override lazy val _table: T = session.records.empty(header).table
+  override lazy val _table: T = session.records.unit(header).table
```
Mats-SX: This seems a bit strange to me. I'd prefer if we had a different call for invoking an empty table of x columns, and separate that from the concept of the unit table (which has no columns by definition).

okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/operators/RelationalOperator.scala
```diff
@@ -133,6 +133,14 @@ abstract class RelationalOperator[T <: Table[T] : TypeTag] extends AbstractTreeN
   }
 }
 
+trait NoRecords[T <: Table[T]] {
```
Mats-SX: Why not call this `UnitRecords`?

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
Mats-SX: I think I prefer the old code where `empty()` and `unit()` were separated. These are separate concepts to me: a table of cardinality 1 and empty column set vs a table of cardinality 0 and non-empty column set.

okapi-relational/src/main/scala/org/opencypher/okapi/relational/api/table/RelationalCypherRecords.scala
```diff
@@ -44,9 +42,7 @@ trait RelationalCypherRecordsFactory[T <: Table[T]] {
 
   type Records <: RelationalCypherRecords[T]
 
-  def unit(): Records
-
-  def empty(initialHeader: RecordHeader = RecordHeader.empty): Records
```
Mats-SX: I prefer to keep these signatures separate.

okapi-logical/src/test/scala/org/opencypher/okapi/logical/impl/logical/LogicalPlannerTest.scala
```diff
@@ -225,10 +231,13 @@ class LogicalPlannerTest extends BaseTestSuite with IrConstruction {
...
           )
         )
-      ),
+      ), irFieldA, irFieldG, irFieldR),
       SolvedQueryModel(
         Set(irFieldA, irFieldG, irFieldR, IRField("a.name")(CTFloat)),
         Set(
+          IsNotNull(varA)(CTBoolean),
+          IsNotNull(varG)(CTBoolean),
+          IsNotNull(varR)(CTBoolean),
```
Mats-SX: It strikes me as a bit odd that we need these filters... There must be something in how we do scans that I do not understand.

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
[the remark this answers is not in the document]
s1ck: @DarthMax and myself had a short discussion about that already. For a node scan that does not start from a driving table, we just [stack the scan on top of the start operator](https://github.com/opencypher/cypher-for-apache-spark/blob/master/okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala#L469-L473). I agree that doing the cross between an empty table and the unit table would resolve the `MATCH` vs `OPTIONAL MATCH` scenario. However, I'm struggling with the second case (no input table). Let's say we plan always plan a cartesian, what would be the inputs in that case?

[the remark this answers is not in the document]
Mats-SX: That stacking on top of Start effectively means that we just _replace_ whatever records that Start fed us with the records we get from the `scanOperator` call on the graph, right? That's as far as I gathered when I studied it during the review. So the stacking is a trick to let the scan mimic a leaf operator really. I suppose that is the source of my confusion.

If we always plan a cartesian (which makes sense to me) and the query starts with `MATCH (:A)`, then the inputs to the cartesian would be the scan table for `(:A)` and the unit table. Crossjoining those would result in exactly the scan table because the unit table turns crossJoin into an identity function (if I understand correctly).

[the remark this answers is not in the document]
s1ck: I guess what you are referring to, is performing a cartesian [here](https://github.com/opencypher/cypher-for-apache-spark/blob/master/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala#L301). The right hand side is the result of a match pattern, e.g. `(n)` which might be empty, i.e. has a header/column names, but 0 rows. The cartesian therefor also produces 0 rows.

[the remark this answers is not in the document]
s1ck: I mean, you're right. Crossjoining some `T |><| UnitTable == T`. But how do we solve the `OPTIONAL MATCH (n) RETURN n` results into `null`. case? If `(n)` returns 0 rows, cross joining it with `UnitTable` results in `0 rows =!= 1 row including null`.
We could - for example - in the case I linked in the previous comment, make sure that the lhs is actually a table containing the null row and has the rhs header. However, consider the following scenario (from Neo4j):

```
CREATE (n:Foo)
OPTIONAL MATCH (n:Bar) RETURN n // returns 1 row: null
OPTIONAL MATCH (n:Foo) RETURN n // returns 1 row: (0:Foo)
```

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
s1ck: I see your point, but `unit(RecordHeader.empty)` gives you exactly what you want.

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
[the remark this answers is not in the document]
Mats-SX: CrossJoin for `MATCH` and OuterJoin for `OPTIONAL MATCH`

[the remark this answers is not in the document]
Mats-SX: So I don't mean crossjoin at the location you pointed to, I mean outer join there and cross join somewhere here:

https://github.com/opencypher/cypher-for-apache-spark/blob/3af20a2beb2230ecb991aa7f92a8b37d819acd90/okapi-relational/src/main/scala/org/opencypher/okapi/relational/impl/planning/RelationalPlanner.scala#L266-L291

In this piece of code `inOp` is used to find a graph, but its records are discarded.

spark-cypher/src/main/scala/org/opencypher/spark/impl/CAPSRecords.scala
Mats-SX: I know, it's just that the concept of the unit table as being a constant is fuzzied. There are many empty tables and there are many tables with 1 row, but there's only one (isomorphic) unit table.

Mats-SX: Why do we need a 1-row table with `null` columns anyway?

Mats-SX: Ohhhh, so _that's_ what this PR does for the `OPTIONAL MATCH`?

s1ck: Exactly :)

s1ck:
Closed in favor of https://github.com/opencypher/cypher-for-apache-spark/pull/858

PR #855, opened 2019-03-06
GraphDDL improvements

s1ck:


s1ck:
Yes, I was gonna add that, too.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -145,97 +145,90 @@ object GraphDdl {
...
+        startNodeType = toNodeType(relationshipTypeDefinition.startNodeType),
+        labels = resolveRelationshipLabel(relationshipTypeDefinition),
+        endNodeType = toNodeType(relationshipTypeDefinition.endNodeType))
+
+    private def resolveNodeTypes(
+      parts: GraphTypeParts,
+      local: PartialGraphType
+    ): Map[NodeTypeDefinition, PropertyKeys] = {
+      parts.nodeTypes
+        .map(nodeType => nodeType.copy(elementTypes = local.resolveNodeLabels(nodeType)))
```
tobias-johansson: Is there a reason that we do `copy` here instead of `NodeTypeDefinition(local.resolveNodeLabels(nodeType))`?
If not, then I think it's more clear to explicitly create new objects

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -145,97 +145,90 @@ object GraphDdl {
...
 
-    private def resolveNodeLabels(nodeType: NodeTypeDefinition): Set[String] =
-      tryWithNode(nodeType.elementTypes)(nodeType.elementTypes.flatMap(resolveElementTypes).map(_.name))
+    private def resolveRelationshipTypes(
+      parts: GraphTypeParts,
+      local: PartialGraphType
+    ): Map[RelationshipTypeDefinition, PropertyKeys] = {
+      parts.relTypes
+        .map(relType => relType.copy(
+          startNodeType = relType.startNodeType.copy(elementTypes = local.resolveNodeLabels(relType.startNodeType)),
```
tobias-johansson: Is there a reason that we do `copy` here instead of 
`RelationshipTypeDefinition()` and `NodeTypeDefinition()`?

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
```diff
@@ -145,97 +145,90 @@ object GraphDdl {
...
 
-  object GraphType {
-    val empty: GraphType = GraphType()
+  object PartialGraphType {
+    val empty: PartialGraphType = PartialGraphType(name = "empty")
   }
 
-  private[graphddl] case class GraphType(
-    parent: Option[GraphType] = None,
+  private[graphddl] case class PartialGraphType(
```
tobias-johansson: Nice!

okapi-testing/src/main/scala/org/opencypher/okapi/testing/PGDSAcceptanceTest.scala
```diff
@@ -119,7 +119,7 @@ trait PGDSAcceptanceTest[Session <: CypherSession, Graph <: PropertyGraph] {
          |CREATE (combo2:A:B { name: 'COMBO2', type: 'AB2' })
          |CREATE (c:C { name: 'C' })
          |CREATE (ac:A:C { name: 'AC' })
-         |CREATE (d { name: 'D', type: 'NO_LABEL' })
+         |//CREATE (d { name: 'D', type: 'NO_LABEL' })
```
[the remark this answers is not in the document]
s1ck: I instead added a second test graph that has a node with empty label set. This test is not run by the SQL PGDS, but by all other PGDSs to not have test coverage regression.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/GraphDdlConversions.scala
```diff
@@ -0,0 +1,185 @@
...
+          val intersect = propertyKeysSets.reduce(_ intersect _)
+          val union = propertyKeysSets.map(_ -- intersect).reduce(_ union _)
+          val unionOptional = union.map { case (key, ct) => key -> ct.nullable }
+          val propertyKeys = intersect ++ unionOptional
+
+          name -> propertyKeys.toMap
+        }
+        .toSeq
+
+      val explicitInheritance = extractInheritance(flatElementTypes)
```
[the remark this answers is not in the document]
s1ck: :) At the moment, it's a cool feature (that we don't use throughout the test code). I wanted to keep it somewhere to continue working on it. 
Producing "flat" types would work and is actually the first step of this algorithm. The `def extractInheritance` tries to recursively move common property keys to the more general type.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
s1ck: Replaced that pattern in all relevant places.

s1ck: I think, I somehow like the `.copy` pattern to make code easier to extend without recompiling and fixing all the constructor calls, but named arguments would work there, too.

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
s1ck: It was charming to build on top of it :)

graph-ddl/src/main/scala/org/opencypher/graphddl/GraphDdl.scala
tobias-johansson: That's a good point :)
Except for the unexpected case when you actually didn't want to propagate other values..

PR #859, opened 2019-03-08
Improve Cypher Value Converters

DarthMax:
Improve `CypherValue.apply` to allow injecting additional converters

### DIFF
--- a/spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
+++ b/spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
@@ -26,11 +26,12 @@
  */
 package org.opencypher.spark.impl
 
+import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
 import org.apache.spark.sql.types._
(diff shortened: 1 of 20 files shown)


spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -497,7 +498,8 @@ object SparkSQLExprMapper {
 
   private def createStructColumn(structColumns: Seq[Column]): Column = {
     if (structColumns.isEmpty) {
-      functions.lit(null).cast(new StructType())
+      val emptyStructUdf = functions.udf( () => new GenericRowWithSchema(Array(), StructType(Nil)), StructType(Nil))
+      emptyStructUdf()
```
soerenreichardt: Why is this better?

DarthMax: the second function actually creates an empty struct, while the old one returned `null` instead. This worked so far, but does not any more (though I forgot why)

PR #860, opened 2019-03-08
Remove CSV support from SQL PGDS

DarthMax:


s1ck:
You can remove [explicit casts](https://github.com/opencypher/cypher-for-apache-spark/blob/master/spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala#L206-L209) now. I introduced those because of CSV. However it's possible that you still need to the `Date` cast, that you did before.

spark-cypher/src/main/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -58,6 +58,11 @@ case class SqlPropertyGraphDataSource(
   idGenerationStrategy: IdGenerationStrategy = SerializedId
 )(implicit val caps: CAPSSession) extends CAPSPropertyGraphDataSource {
 
+  require(
+    sqlDataSourceConfigs.forall {case (_, config) => config.format != FileFormat.csv},
```
s1ck:
```suggestion
    sqlDataSourceConfigs.forall { case (_, config) => config.format != FileFormat.csv },
```

PR #865, opened 2019-03-13
ListSlice for non-negative and valid ranges

pstutz:


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -917,6 +925,22 @@ final case class ExistsPatternExpr(targetField: Var, ir: CypherQuery)
...
+      Some(newChildren.head) -> newChildren.tail
+    } else {
+      None -> newChildren
+    }
+    val indexed = as.zipWithIndex
+    val conditions = indexed.collect { case (c, i) if i % 2 == 0 => c}
+    val values = indexed.collect { case (c, i) if i % 2 == 1 => c}
+    val newAlternatives = conditions.zip(values).toList
+    CaseExpr(newAlternatives, newDefault)(cypherType)
+  }
```
Mats-SX: I suppose this is necessary because the `okapi-trees` do not do well with lists of tuples? Or is there another reason?

spark-cypher-tck/src/test/resources/failing_blacklist
```diff
@@ -153,7 +153,6 @@ Feature "MatchAcceptance": Scenario "Return relationships by collecting them as
 Feature "MatchAcceptance": Scenario "Return a var length path"
 Feature "MatchAcceptance": Scenario "Return a var length path of length zero"
 Feature "MatchAcceptance": Scenario "Return a named var length path of length zero"
-Feature "MatchAcceptance": Scenario "Accept skip zero"
```
Mats-SX: I wonder what caused this to work now...

spark-cypher-tck/src/test/resources/failing_blacklist
```diff
@@ -245,19 +240,8 @@ Feature "ListOperations": Scenario "IN should return true if correct list found
...
-Feature "ListOperations": Scenario "List slice with implicit start"
-Feature "ListOperations": Scenario "List slice with empty range"
 Feature "ListOperations": Scenario "List slice with invalid range"
-Feature "ListOperations": Scenario "List slice with parameterised range"
-Feature "ListOperations": Scenario "List slice with null range #1"
-Feature "ListOperations": Scenario "List slice with null range #2"
-Feature "ListOperations": Scenario "List slice with null range #3"
-Feature "ListOperations": Scenario "List slice with null range #4"
-Feature "ListOperations": Scenario "List slice with null range #5"
-Feature "ListOperations": Scenario "List slice with implicit end"
```
Mats-SX: Good stuff

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
```diff
@@ -176,210 +179,59 @@ class LogicalPlanner(producer: LogicalOperatorProducer)
...
-        val plannedInner = inner.foldLeft(in)((op, expr) => planInnerExpr(expr, op))
-        producer.projectExpr(ors, plannedInner)
+        producer.planExistsSubQuery(ex, in, this (ex.ir))
 
-      case containerIndex@ ContainerIndex(container, idx) =>
-        val withContainer = planInnerExpr(container, in)
-        val withIndex = planInnerExpr(idx, withContainer)
-        producer.projectExpr(containerIndex, withIndex)
+      case _ =>
+        expr.children.foldLeft(in)((op, e) => planInnerExpr(e, op))
```
Mats-SX: I don't really get what the actual operation is here; where is the base case? It seems as though we only do anything for `ExistsPatternExpr`, and for everything else we recursively traverse to the leaves of the `Expr` tree and at that point just return the input plan. So if `ExistsPatternExpr` is not part of the tree, we will traverse to the bottom and back up again returning the initial input plan, unchanged. There's no `producer.planXYZ()` call on the recursive path.

In the old code this was true for some cases and not true for some other cases; this code seems to claim it is entirely unnecessary except for subqueries.

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/FunctionTests.scala
```diff
@@ -1630,6 +1630,31 @@ class FunctionTests extends CAPSTestSuite with ScanGraphInit {
     }
   }
 
+  describe("list slice") {
```
Mats-SX: I think these should move to `ExpressionTests` because list slice is not a function.

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -352,6 +343,10 @@ object SparkSQLExprMapper {
             case other => throw NotImplementedException(s"Accessing $other by index is not supported")
           }
 
+        case ListSlice(_, maybeFrom, maybeTo) =>
+          require(maybeFrom.isDefined || maybeTo.isDefined)
```
Mats-SX: This feels like a requirement better checked at `ExpressionConverter` time, no? It's essentially a signature validation.

Mats-SX: In _this_ case we could compile into specific `ListSlice` expressions, making this code less overloaded:

```
case class ListSliceFrom(expr: Expr, from: Expr) => list_slice(child0, Some(child1))
case class ListSliceFromTo(expr: Expr, from: Expr, to: Expr) => list_slice(child0, Some(child1), Some(child2))
case class ListSliceTo(expr: Expr, to: Expr) => list_slice(child0, None, Some(child2))
```

(not valid code but you get the point)

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
pstutz: Yes, automated child detection for lists of tuples is not yet implemented.

spark-cypher-tck/src/test/resources/failing_blacklist
pstutz: Maybe consistency in the planner. :P

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
pstutz: I'm happy to do more cleanups. :) I assume we should be able to express the whole thing as a simple tree traversal, with the expressions unchanged. Maybe that will require converting the `exists subquery` into a non-expression sooner.

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/FunctionTests.scala
pstutz: Done

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
pstutz: Done

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
Mats-SX: Doesn't really answer my question :) There are many calls to `produce.planExpr()` in the red code that no longer exist. This means that we do not plan `Project()` plan operators anymore; isn't that strange?

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -343,9 +343,9 @@ object SparkSQLExprMapper {
             case other => throw NotImplementedException(s"Accessing $other by index is not supported")
           }
 
-        case ListSlice(_, maybeFrom, maybeTo) =>
-          require(maybeFrom.isDefined || maybeTo.isDefined)
-          list_slice(child0, if (maybeFrom.isDefined) Some(child1) else None, if (maybeFrom.isDefined) convertedChildren.lift(2) else Some(child1))
+        case _: ListSliceFromTo => list_slice(child0, Some(child1), Some(child2))
+        case _: ListSliceFrom => list_slice(child0, Some(child1), None)
+        case _: ListSliceTo => list_slice(child0, None, Some(child1))
```
Mats-SX: Nice!

okapi-logical/src/main/scala/org/opencypher/okapi/logical/impl/LogicalPlanner.scala
pstutz: I think they are all captured by the generic `expr.children.foldLeft(in)((op, e) => planInnerExpr(e, op))` call.

PR #868, opened 2019-03-13
Fix ConsoleApp trait name

pstutz:
Should fix master build that broke after bringing the Neo4j tests back.


pstutz:
Merging this without review, as I think that a broken master is worse than the risk that this might not fix it.

PR #870, opened 2019-03-14
Support creation of big decimal

soerenreichardt:


Mats-SX:
Spark outputs doubles when doing stuff like `BigDecimal + Float` which I find rather strange; the wider type can be used, but isn't. This results in rounding errors when combining these types... which is the whole reason for using a BigDecimal in the first place, one would think.

I've went with mimicking Spark here, but I wonder if it is sane. Perhaps we should just disallow float and BigDecimal arithmetics altogether, because we couldn't just change it without breakage.

jjaderberg:
@Mats-SX How about opening a jira ticket with spark (or GH issue or e-mail the dev list)?
When spark behaves oddly, I think we should ask spark wtf. If there is a good reason, we learn; if there isn't, we can work around with confidence (or fix spark if within scope). In any case, asking is nice and knowing reasons is useful.

Mats-SX:
@jjaderberg Good idea: https://issues.apache.org/jira/projects/SPARK/issues/SPARK-27283

I issued it as a bug, because I tried something equivalent in MySQL and the same behaviour does not seem to occur there.

spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/BigDecimalTests.scala
```diff
@@ -46,59 +46,59 @@ class BigDecimalTests extends CAPSTestSuite with ScanGraphInit {
...
         )
       )
     }
 
     it("subtracts a big decimal and a float") {
-      caps.cypher("RETURN bigdecimal(1234, 2) - 10.2 AS decimal").records.toMaps should equal(
+      caps.cypher("RETURN bigdecimal(1234, 4, 2) - 10.2 AS decimal").records.toMaps should equal(
         Bag(
-          CypherMap("decimal" -> 2.1400000000000006) // this is crazy! this is why it should be BigDecimal and not float
+          CypherMap("decimal" -> 2.1400000000000006)
```
Mats-SX: I still think this is kinda weird x)

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/ExpressionConverter.scala
```diff
@@ -53,28 +53,32 @@ object AddType {
...
+        case _ => None
+      }
+
+      result.map(_.asNullableAs(rhs join lhs))
     }
     val sigs = Set(addSignature, BigDecimalSignatures.arithmeticSignature(Addition))
 
-    returnTypeFor(Seq.empty, Seq(lhs, rhs), sigs)
+    val foo = returnTypeFor(Seq.empty, Seq(lhs, rhs), sigs)
+    foo
```
Mats-SX: Can inline `foo`

spark-cypher-testing/src/test/scala/org/opencypher/spark/api/io/sql/SqlPropertyGraphDataSourceTest.scala
```diff
@@ -105,68 +107,40 @@ class SqlPropertyGraphDataSourceTest extends CAPSTestSuite with HiveFixture with
...
     )
 
-    it("reads bigdecimals as standard properties") {
+    def testBigDecimalConversion(sourcePrecision: Int, sourceScale: Int, targetPrecision: Int, targetScale: Int): Unit = {
+      val sourceValue = BigDecimal(1*Math.pow(10, sourcePrecision).toInt, sourceScale)
+      val sourceType = DecimalType(sourcePrecision, sourceScale)
 
-      val value = BigDecimal(314159, 5)
-      val typ = DecimalType(10, 5)
+      val targetValue = sourceValue.setScale(targetScale, RoundingMode.HALF_UP)
```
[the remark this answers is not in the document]
DarthMax: We actually coerce the values to the user defined type. I think this is the correct behavior as this is the type the user expects.

PR #871, opened 2019-03-14
Rewrite neo4j launcher tasks

tobias-johansson:
- Enterprise fails if instances > 1, so launch Community in that case
- Extract tasks into method for code sharing
- Add status task

### DIFF
--- a/okapi-neo4j-io-testing/src/services/scala/org/opencypher/testing/services/Neo4jLauncher.scala
+++ b/okapi-neo4j-io-testing/src/services/scala/org/opencypher/testing/services/Neo4jLauncher.scala
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2016-2019 "Neo4j Sweden, AB" [https://neo4j.com]
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


okapi-neo4j-io-testing/src/services/scala/org/opencypher/testing/services/Neo4jLauncher.scala
```diff
@@ -0,0 +1,52 @@
...
+trait Neo4jLauncher {
+
+  def builder: TestServerBuilder
+
+  def main(args: Array[String]): Unit = {
+    val Array(marker, instances) = args
+
+    Range(0, instances.toInt).foreach { i =>
+      val bolt = 7687 + i
+      val http = 7474 + i
```
Mats-SX: Can we use other ports than the built-in default? I usually have a Neo4j instance running locally and I think this may be in conflict?

tobias-johansson: We could certainly change this. Do you think the examples that connect to neo4j should also have non-standard ports?

Mats-SX: I don't care so much about those; they can use non-default or default, so whichever seems easiest.

PR #875, opened 2019-04-02
Address TODO items in SparkSQLExprMapper

pstutz:
I addressed the reviews on a branch that is based on the union types branch to avoid another merge. Will move the changes over once we know what will happen to union types.


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -438,19 +438,72 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
 
 }
 
-final case class Property(entity: Expr, key: PropertyKey)(val cypherType: CypherType) extends Expr {
+sealed trait Property extends Expr {
+
+  def entity: Expr
```
Mats-SX: Should we perhaps call this `owner` so that it is more generic? Alternatively `container`, `object`, `struct`... but I don't like any of those.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -438,19 +438,72 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
...
 
+final case class EntityProperty(entity: Expr, key: PropertyKey)(val cypherType: CypherType) extends Property {
+
+  override def withOwner(v: Var): EntityProperty = EntityProperty(v, key)(cypherType)
+
+}
+
+final case class MapProperty(map: Expr, key: PropertyKey) extends Property {
+
+  override def entity: Expr = map
```
Mats-SX: If we didn't call the interface field `entity` it would probably feel less awkward to just `case class`-override it directly from the trait. I feel like the field name `map` (and below `date`, `duration` etc) doesn't add that much since the class name still tells us something about the owner.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -438,19 +438,72 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
...
+
+  val cypherType: CypherType = map.cypherType match {
+    case CTMap(inner) =>
+      inner.getOrElse(key.name, CTVoid)
```
Mats-SX: I think it is `CTNull` if the key doesn't exist.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -438,19 +438,72 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
...
+
+  override def withOwner(v: Var): MapProperty = MapProperty(v, key)
+
+}
+
+final case class DateProperty(date: Expr, key: PropertyKey) extends Property {
+
+  override def entity: Expr = date
+
+  val cypherType: CypherType = CTInteger.asNullableAs(date.cypherType)
```
Mats-SX: This is nice!

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/ExpressionConverter.scala
```diff
@@ -102,26 +102,33 @@ final class ExpressionConverter(context: IRBuilderContext) {
       ListLit(elements)(CTList(elementType))
 
     case ast.Property(m, ast.PropertyKeyName(name)) =>
-      val mapLike = convert(m)
-      val propertyType = mapLike.cypherType.material match {
-        case CTVoid => CTNull
+      val entity = convert(m)
```
Mats-SX: Could be `owner` re my comment above

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/ExpressionConverterTest.scala
```diff
@@ -286,7 +286,7 @@ class ExpressionConverterTest extends BaseTestSuite with Neo4jAstTestSupport {
 
   it("can convert property access") {
     convert(prop("NODE", "age")) shouldEqual
-      Property('NODE, PropertyKey("age"))(CTInteger)
+      EntityProperty('NODE, PropertyKey("age"))(CTInteger)
```
Mats-SX: I think it would be nice with some tests that make sure the other forms of `Property` get created in the right circumstances, and that they get the right type (`CTInteger` for dates etc).

spark-cypher/src/main/scala/org/opencypher/spark/impl/SparkSQLExprMapper.scala
```diff
@@ -116,20 +116,12 @@ object SparkSQLExprMapper {
...
-                NULL_LIT
-              } else {
-                column_for(expr)
-              }
-          }
+        case _: EntityProperty => if (!header.contains(expr)) NULL_LIT else column_for(expr)
+        case MapProperty(_, key) => if (expr.cypherType.material == CTVoid) NULL_LIT else child0.getField(key.name)
+        case DateProperty(_, key) => temporalAccessor[java.sql.Date](child0, key.name)
+        case LocalDateTimeProperty(_, key) => temporalAccessor[java.sql.Timestamp](child0, key.name)
+        case DurationProperty(_, key) => TemporalUdfs.durationAccessor(key.name.toLowerCase).apply(child0)
```
Mats-SX: This is much nicer I think -- great win.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -438,19 +438,72 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
 
 }
 
-final case class Property(entity: Expr, key: PropertyKey)(val cypherType: CypherType) extends Expr {
+sealed trait Property extends Expr {
```
Mats-SX: For completeness, but more an okapi concern than a CAPS one, we need to add a generic `Property` implementation as well, for when the owner type is unknown. This isn't supported in CAPS but may be supported by e.g. mem-cypher.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
pstutz: `owner` is already defined on `Expr`  to get the owner of an alias. Going with `propertyOwner`.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
pstutz: I started working on this, but then came to the conclusion that it is better to drive this work with a concrete implementation that can handle such properties. In terms of functionality it is not a regression, as it would have failed during expression conversion before and it still fails the same way.

pstutz:
If union types are merged, then this PR is replaced by https://github.com/opencypher/cypher-for-apache-spark/pull/884

pstutz:
Comments are addressed here: https://github.com/opencypher/cypher-for-apache-spark/pull/884
Motivation: This one was rebased on the union types PR and avoids additional merges, as the modified classes have a lot of overlap.

PR #881, opened 2019-04-05
Yelp demo improvement

FlorentinD:
Fixes queries in Part 4 (WHERE Clause bug), adds some checks if graphs already constructed and integrates tip.json into property graph.


FlorentinD:
Also adds analysis based on edge-count in part 4, which also correlates highly (even higher than page-rank).
Therefore the page-rank based analysis may be not relevant anymore for part 4?! (maybe another use-case?)

s1ck:
Maybe we could also brand Part 4 as "Reverse engineering elite status, is it page rank or just average degree?" the p-value could give us an answer.

PR #883, opened 2019-04-08
Minor things for training / summit

s1ck:


spark-cypher-examples/src/main/scala/org/opencypher/spark/integration/yelp/Part2_YelpGraphLibrary.scala
```diff
@@ -86,8 +88,28 @@ object Part2_YelpGraphLibrary extends App {
...
-         |    CREATE (user1)-[:CO_REVIEWS]->(user2)
+         |    CREATE (user1)-[r:CO_REVIEWS { $reviewCountProperty : $reviewCountProperty }]->(user2)
+         |  RETURN GRAPH
+         |}
+     """.stripMargin)
+
+    // Compute (u1:User)-[:CO_REVIEWS]->(u2:User)-[:REVIEWS]->(:Business)<-[:REVIEWS]-(u1) graph
+    cypher(
+      s"""
+         |CATALOG CREATE GRAPH $fsNamespace.${coReviewAndBusinessGraphName(year)} {
```
jjaderberg: using variables to avoiding duplicating literal values may be good codez, but it also makes the queries much harder to read.

PR #891, opened 2019-04-14
Unignore working tests

FlorentinD:
### Summary of reviewing ignored tests: 
* unchanged ignored tests have been left out of this summary
#### okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
* 902:    ignore("fails when setting a label on an entity that is not in scope") {
    * succeeds more or less: only another error thrown UnTypedExpr instead of UnsupportedOperationException
* 915:    ignore("fails when setting a label on a relationship that is out of scope") {
    * succeeds now 

#### spark-cypher-tck/src/test/scala/org/opencypher/spark/testing/TckSparkCypherTest.scala
* 109:  ignore("run custom scenario") {
     * removeable as not used?
* 118:  ignore("run single scenario") {
     * removeable as now acceptance-test generator for this purpose?
  
#### spark-cypher-testing/src/test/scala/org/opencypher/spark/impl/acceptance/UnwindBehaviour.scala
* 113:  ignore("unwind from expression with empty and null lists") {
    *  _!_ now throws other error than stated in todo above 
(needs more time to investigate, if this changes anything)


okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -913,11 +913,10 @@ class IrBuilderTest extends IrTestSuite {
           |  SET a :Label
           |RETURN GRAPH""".stripMargin
 
-      intercept[UnsupportedOperationException](query.asCypherQuery().model)
+      intercept[UnTypedExpr](query.asCypherQuery().model) //todo: also correct to throw UnTypedExpr instead of UnsupportedOperationException?
```
Mats-SX: I think this exception is perfectly acceptable. It means that it looks for the type of `a` but can not find it, which seems correct.
```suggestion
      intercept[UnTypedExpr](query.asCypherQuery().model)
```

spark-cypher-tck/src/test/scala/org/opencypher/spark/testing/TckSparkCypherTest.scala
```diff
@@ -131,7 +131,8 @@ class TckSparkCypherTest extends CAPSTestSuite {
 
   }
 
-  ignore("run custom scenario") {
+  //todo: remove as no customtest?
```
Mats-SX: I agree we can remove it. It isn't hard to get it back if we want to write TCK-style tests in the future.

PR #892, opened 2019-04-14
Adjust AcceptanceTest Generator to new TCK-API version

FlorentinD:
Due to the updated TCK API dependency, some generated Test couldn't not compile anymore. (Problem with empty constructors)
The solution improves the creation of construct-strings for TCKCypherNode and TCKCypherRelationship. 
For instance node with no labels and no properties now leads to _TCKCypherNode()_ instead of _TCKCypherNode(Set(), TCKCypherPropertyMap(Map()))_

### DIFF
--- a/okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/CreateStringGenerator.scala
+++ b/okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/CreateStringGenerator.scala
@@ -51,8 +51,19 @@ object CreateStringGenerator {
           case _ => s"TCKCypherValue.apply(${escapeString(l.toString)}, false)"
         }
       case TCKCypherNull => "TCKCypherNull"
-      case TCKCypherNode(labels, properties) => s"TCKCypherNode(Set(${labels.map(escapeString).mkString(",")}), ${tckCypherValueToCreateString(properties)})"
-      case TCKCypherRelationship(typ, properties) => s"TCKCypherRelationship(${escapeString(typ)}, ${tckCypherValueToCreateString(properties)})"
+      case TCKCypherNode(labels, properties) =>
+        val labelsString = if (labels.isEmpty) "" else s"Set(${labels.map(escapeString).mkString(",")})"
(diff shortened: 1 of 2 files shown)


okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/CreateStringGenerator.scala
```diff
@@ -51,8 +51,19 @@ object CreateStringGenerator {
...
-      case TCKCypherNode(labels, properties) => s"TCKCypherNode(Set(${labels.map(escapeString).mkString(",")}), ${tckCypherValueToCreateString(properties)})"
-      case TCKCypherRelationship(typ, properties) => s"TCKCypherRelationship(${escapeString(typ)}, ${tckCypherValueToCreateString(properties)})"
+      case TCKCypherNode(labels, properties) =>
+        val labelsString = if (labels.isEmpty) "" else s"Set(${labels.map(escapeString).mkString(",")})"
+        val propertyString = if (properties.properties.isEmpty) "" else s"${tckCypherValueToCreateString(properties)}"
+        val separatorString =
+          if (labels.isEmpty && properties.properties.isEmpty) ""
+          else if (labels.isEmpty) "properties = "
+          else if (properties.properties.isEmpty) ""
+          else ", "
```
Mats-SX: It's empty string if `properties.properties.isEmpty` regardless of `labels` here; we could probably simplify this somewhat:

```
          if (properties.properties.isEmpty) ""
          else if (labels.isEmpty) "properties = "
          else ", "
```

s1ck:
@FlorentinD I was wondering if you test the correctness of the code generation. If the TCK dependency changes in a way that the generated code does not compile anymore, it should be detected by our tests.

FlorentinD:
@s1ck currently i run the generated tests manually.
The "problem" is, that after the gradle task is run, the generated tests won't be executed by the CI. 
On the other hand do we want to execute all TCK-scenarios twice in CI? 

Maybe it would be possible to check, whether they can be compiled in CI

PR #885, opened 2019-04-08
Make more expressions self typing

DarthMax:

PR #894, opened 2019-04-16
Temporal aggregation tests

FlorentinD:
Based on #885 

Creating these tests led to following problems:
* Duration can't be aggregated as the spark type calendarinterval is not supported for aggregation 
  -> will be looked into in the future
- [x] Avg & Sum over DateTimes and Dates is allowed in spark, but not in cypher  (error now thrown)
- [x] Spark & Cypher don't support f.i. Avg over Dates, but the error isn't catched in CAPS yet (error now caught)


FlorentinD:
Type-checking for other Expr will follow in an extra PR

PR #897, opened 2019-04-18
Add draft for binary switch based on extended CAPSSession

s1ck:
Extends `CAPSSession` with SPIP API via mixin trait. The trait and necessary adapters are bundled within a single package. This allows a seamless switch between SPIP and OKAPI API in a single graph app by switching the session implementation. See `GraphApp` for an example.

Pro:
* no additional module
* seamless binary switch between CAPS and SPIP
* CAPS spark-cypher can be renamed to `morpheus-spark-cypher` or `morpheus`

Cons:
* as with v1, it feels a bit weird that similar named methods return equally named results but with different semantics, but I guess that's what we just need to deal with

Needs apache/spark#24297 installed locally to compile and run.


Mats-SX:
I think I like this one better than the alternative, but I'm not sure I understand all of the implications. Very nice examples though!

PR #905, opened 2019-04-30
Expression Type Checking

FlorentinD:
Adds checks, whether the cypher-types of the argument-expressions for an expression are valid.
Thus f.i. `Trim(Var('e')(CTInteger)).cyphertype` will now fail.

Also some redundancy could get eliminated in the process

### DIFF
--- a/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
+++ b/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
@@ -179,13 +179,18 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
     it("compares less than between different types") {
 
       // Given
-      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+      val given = initGraph(
+        """CREATE (:A {val: 4})-[:REL]->(:B {val2: 1.0}),
+          |       (:A {val: 1})-[:REL]->(:B {val2: 4.0})
+        """.stripMargin)
 
       // When
       val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val < b.val2 RETURN a.val")
 
       // Then
-      result.records.collect.toBag shouldBe empty
+      result.records.collect.toBag should equal(Bag(
+        CypherMap("a.val" -> 1)
+      ))
     }
 
     it("less than or equal") {
@@ -209,13 +214,19 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
     it("compares less than or equal between different types") {
 
       // Given
-      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+      val given = initGraph(
+        """CREATE (:A {val: 4})-[:REL]->(:B {val2: 4.0}),
+          |       (:A {val: 1})-[:REL]->(:B {val2: 4.0})
+        """.stripMargin)
 
       // When
       val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val <= b.val2 RETURN a.val")
 
       // Then
-      result.records.collect.toBag shouldBe empty
+      result.records.collect.toBag should equal(Bag(
+        CypherMap("a.val" -> 4),
+        CypherMap("a.val" -> 1)
+      ))
     }
 
     it("greater than") {
@@ -234,13 +245,18 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
     it("compares greater than between different types") {
 
       // Given
-      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+      val given = initGraph(
+        """CREATE (:A {val: 4})-[:REL]->(:B {val2: 1.0}),
+          |       (:A {val: 1})-[:REL]->(:B {val2: 4.0})
+        """.stripMargin)
 
       // When
       val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val > b.val2 RETURN a.val")
 
       // Then
-      result.records.collect.toBag shouldBe empty
+      result.records.collect.toBag should equal(Bag(
+        CypherMap("a.val" -> 4)
+      ))
     }
 
     it("greater than or equal") {
@@ -260,13 +276,19 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
     it("compares greater than or equal between different types") {
 
       // Given
-      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+      val given = initGraph(
+        """CREATE (:A {val: 4})-[:REL]->(:B {val2: 1.0}),
+          |       (:A {val: 4})-[:REL]->(:B {val2: 4.0})
+        """.stripMargin)
 
       // When
       val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val >= b.val2 RETURN a.val")
 
       // Then
-      result.records.collect.toBag shouldBe empty
+      result.records.collect.toBag should equal(Bag(
+        CypherMap("a.val" -> 4),
+        CypherMap("a.val" -> 4)
+      ))
     }
   }
 
--- a/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
+++ b/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
@@ -199,8 +199,11 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       // Given
       val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
 
+      // Where
+      val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val < b.val2 RETURN a.val")
+
       // Then
-      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val < b.val2 RETURN a.val"))
+      result.records.collect.toBag shouldBe empty
     }
 
     it("less than or equal") {
@@ -244,8 +247,11 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       // Given
       val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
 
+      // Where
+      val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val <= b.val2 RETURN a.val")
+
       // Then
-      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val <= b.val2 RETURN a.val"))
+      result.records.collect.toBag shouldBe empty
     }
 
     it("greater than") {
@@ -283,8 +289,11 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       // Given
       val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
 
+      // Where
+      val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val > b.val2 RETURN a.val")
+
       // Then
-      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val > b.val2 RETURN a.val"))
+      result.records.collect.toBag shouldBe empty
     }
 
     it("greater than or equal") {
@@ -324,8 +333,11 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       // Given
       val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
 
+      // Where
+      val result = given.cypher("MATCH (a:A)-->(b:B) WHERE a.val >= b.val2 RETURN a.val")
+
       // Then
-      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val >= b.val2 RETURN a.val"))
+      result.records.collect.toBag shouldBe empty
     }
   }
 
--- a/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
+++ b/morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
@@ -28,6 +28,7 @@ package org.opencypher.morpheus.impl.acceptance
 
 import org.opencypher.morpheus.testing.MorpheusTestSuite
 import org.opencypher.okapi.api.value.CypherValue._
+import org.opencypher.okapi.impl.exception.NoSuitableSignatureForExpr
 import org.opencypher.okapi.testing.Bag
 import org.opencypher.okapi.testing.Bag._
 
@@ -193,6 +194,15 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       ))
     }
 
+    it("fails when comparing less than between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val < b.val2 RETURN a.val"))
+    }
+
     it("less than or equal") {
       // Given
       val given = initGraph(
@@ -229,6 +239,15 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       ))
     }
 
+    it("fails when comparing less than or equal between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val <= b.val2 RETURN a.val"))
+    }
+
     it("greater than") {
       // Given
       val given = initGraph("""CREATE (:Node {val: 4})-[:REL]->(:Node {val: 5})""")
@@ -257,6 +276,15 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
       result.records.collect.toBag should equal(Bag(
         CypherMap("a.val" -> 4)
       ))
+    }
+
+    it("fails when comparing greater than between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val > b.val2 RETURN a.val"))
     }
 
     it("greater than or equal") {
@@ -289,6 +317,15 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
         CypherMap("a.val" -> 4),
         CypherMap("a.val" -> 4)
       ))
+    }
+
+    it("fails when comparing greater than or equal between different types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val >= b.val2 RETURN a.val"))
     }
   }
 
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
@@ -631,15 +631,8 @@ sealed trait FunctionExpr extends Expr {
          case None => typ
         }
       case None =>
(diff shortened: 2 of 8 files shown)


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
 }
 
 // Functions
-sealed trait FunctionExpr extends Expr {
+sealed trait PropagationTyp
+final case class ChildNullPropagation() extends PropagationTyp
```
DarthMax: You can make all of these `case objects`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+sealed trait FunctionExpr extends Expr {
   override final def toString = s"$name(${exprs.mkString(", ")})"
   override final def withoutType = s"$name(${exprs.map(_.withoutType).mkString(", ")})"
   def name: String = this.getClass.getSimpleName.toLowerCase
   def exprs: List[Expr]
-}
 
-final case class MonotonicallyIncreasingId() extends FunctionExpr {
+  protected def cypherType(typePropagationFunction: Option[PropagationTyp] = None): CypherType = {
+    val materialTypeVector = exprs.foldLeft(Vector.empty[CypherType]) {
```
DarthMax: I think you can simplify this to `expr.map(_.cypherType.material)`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+      case (acc , e: Expr) => acc :+ e.cypherType.material
+    }
+    val joinedCypherType = exprs match {
```
DarthMax: Here you could use`exprs.map(_.cypherType).foldLeft(CTVoid)(_ join _)`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+    }
+    val maybeType = signature(materialTypeVector)
 
-  override def exprs: List[Expr] = List.empty
+    maybeType match {
+      case Some(typ) =>
+       typePropagationFunction match {
+         case Some(_: ChildNullPropagation) => childNullPropagatesTo(typ)
+         case Some(_: NullabilityPropagation) => typ.asNullableAs(joinedCypherType)
+         case Some(_: FullNullabilityPropagation) =>  if (exprs.exists(!_.cypherType.isNullable)) typ else typ.nullable
```
DarthMax: I think this one and the one above do the same thing. If one of the types is nullable joinedCypherType will be nullable

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
 
+sealed trait FunctionExpr extends Expr {
   override final def toString = s"$name(${exprs.mkString(", ")})"
   override final def withoutType = s"$name(${exprs.map(_.withoutType).mkString(", ")})"
   def name: String = this.getClass.getSimpleName.toLowerCase
   def exprs: List[Expr]
-}
 
-final case class MonotonicallyIncreasingId() extends FunctionExpr {
+  protected def cypherType(typePropagationFunction: Option[PropagationTyp] = None): CypherType = {
```
DarthMax: maybre rename this to `computeCypherType` to avoid confusion with the `val`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+         case None => typ
+        }
+      case None =>
+          if (children.exists(_.cypherType == CTNull)) CTNull //todo: check corner cases (cases where no Propagation is specified)
+          else throw UnsupportedOperationException(s"Type signature ${getClass.getSimpleName}($joinedCypherType) is not supported.")
+    }
+  }
 
-  override def cypherType: CypherType = CTInteger
+  def signature(inputCypherTypes: Vector[CypherType]): Option[CypherType]
```
DarthMax: Is there a specific reason you are using `vector` here. If not, we usually use `Seq`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+  }
 }
 
 object PrefixId {
   type GraphIdPrefix = Byte
 }
 
 final case class PrefixId(expr: Expr, prefix: GraphIdPrefix) extends UnaryFunctionExpr {
-  override def cypherType: CypherType = childNullPropagatesTo(CTIdentity)
+  override def cypherType: CypherType = cypherType(Some(ChildNullPropagation()))
```
DarthMax: Instead of doing it like this you could have a default implementation of 
```
override def cypherType: CypherType = cypherType(PropagationType)
```
in FunctionExpr and have the children implement something like
```
def propagationType: Option[PropagationType]
```

that way you don't have to calll cypherType(...) every time

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -600,91 +600,194 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
 }
 
 // Functions
-sealed trait FunctionExpr extends Expr {
+sealed trait PropagationTyp
```
Mats-SX: Should we perhaps call it `PropagationType`?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/parse/functions/FunctionExtensions.scala
```diff
@@ -26,124 +26,43 @@
...
-case object ToBoolean extends Function with TypeSignatures {
-  override val name = functions.ToBoolean.name
-
-  override val signatures = Vector(
-    TypeSignature(argumentTypes = Vector(CTString), outputType = CTBoolean),
-    TypeSignature(argumentTypes = Vector(CTBoolean), outputType = CTBoolean)
-  )
-}
-
-case object ToString extends Function with TypeSignatures {
```
Mats-SX: It's super nice that you managed to remove these, but I don't understand how you were able to without changes also in `ExpressionConverter`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: I knew there had to be something like this 👍

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: yes in combination with using Seq instead of Vector this should work

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: _FullNullabilityPropagation_ will lead to nullability only if *all* children are nullable.
Maybe I can rewrite the condition, making it more readable.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: I used `vector` as it is used for the signatures at the frontend, but sure I can use `Seq`.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/parse/functions/FunctionExtensions.scala
FlorentinD: Currently in the `ExpressionConverter` the signatures are not used (only the name gets checked for the ones the frontend does not know). 
`ToBoolean` f.i. is not used at all, as it just gets converted into the Expr without checks.
 The tests are still working, so I dont think, I missed something there 😅

FlorentinD: I think they were used before, but were forgotten to delete after some refactoring regarding the typing.

FlorentinD:
Currently it only throws an error when `expr.cyphertype` is called.
Maybe an error could already be thrown when constructing an invalid expr?

Mats-SX:
If the field is a `val` it should indeed throw already at object creation. If it is `def` it needs the explicit call. Did you see it behave differently?

I'd say that using `val` and not having an explicit call is a bit nicer.

FlorentinD:
@Mats-SX I encountered some NullPointer-Exceptions using `val`, if will look further into it

Mats-SX:
Right, so sometimes when using `val`s that are referencing other classes' `val`s (in nested calls) you could end up with order of evaluation issues, where some fields have not been initialised before the reference occurs.

FlorentinD:
this seems about right. The `propagationType` is not initialized at this moment

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -334,104 +355,108 @@ final case class HasType(rel: Expr, relType: RelType) extends PredicateExpressio
...
 // Binary expressions
+sealed trait BinaryStrExprSignature {
+  def signature(lhsType : CypherType, rhsType : CypherType): Option[CypherType] = lhsType -> rhsType match {
+    case (CTString, CTString) => Some(CTBoolean)
+    case _ => None
+  }
+}
 
-sealed trait BinaryExpr extends Expr {
+sealed trait BinaryInEqualityExprSignature {
```
jjaderberg: `InEquality` => `Inequality`
Don't need `Binary` -- suggest `InequalityExprSignature`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -334,104 +355,108 @@ final case class HasType(rel: Expr, relType: RelType) extends PredicateExpressio
...
+    case u : CTUnion if u.alternatives.forall(_.subTypeOf(CTRelationship)) => Some(CTBoolean)
+    case _ => None
+  }
 }
 
 final case class IsNull(expr: Expr) extends PredicateExpression {
-
-  override val cypherType: CypherType = CTBoolean
   def inner: Expr = expr
+  override val cypherType: CypherType = CTBoolean
```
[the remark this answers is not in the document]
FlorentinD: seems reasonable to me :)

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -117,41 +113,72 @@ object Var {
   }
 }
 
-final case class ListSegment(index: Int, listVar: Var) extends Var {
+sealed trait PropagationType
+case object ChildNullPropagation extends PropagationType
+case object NullabilityPropagation extends PropagationType
+case object FullNullabilityPropagation extends PropagationType
+case object InPropagation extends PropagationType //as In has no uniform PropagationType
```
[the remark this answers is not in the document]
FlorentinD: Yes I think using `override` instead of `InPropagation` makes sense, I didn't inline it until now, because I need to define the signature nonetheless and something like 
`def signature = None` seemed intuitive to me, and `def signature = ???` even worse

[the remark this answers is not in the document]
FlorentinD: And inlining the logic will be hard, as `childNullPropagatesTo` has the implicit argument `this` in contrast to the other 2 as you stated (excluding `InPropagation`).

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -117,41 +113,60 @@ object Var {
...
+    maybeType match {
+      case Some(typ) =>
+        propagationType match {
+          case Some(NullOrAnyNullable) => childNullPropagatesTo(typ)
+          case Some(AnyNullable) => if(exprs.exists(_.cypherType.isNullable)) typ.nullable else typ
+          case Some(AllNullable) =>  if (exprs.forall(_.cypherType.isNullable)) typ.nullable else typ
+          case None => typ
+        }
+      case None =>
+        if (children.exists(_.cypherType == CTNull)) CTNull //todo: check corner cases (cases where no Propagation is specified)
```
Mats-SX:
```suggestion
        if (children.exists(_.cypherType == CTNull)) CTNull
```

morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
```diff
@@ -179,13 +180,27 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
...
+      ))
+    }
+
+    it("fails when comparing less than between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val < b.val2 RETURN a.val"))
```
Mats-SX: This should not throw exception, it should return an empty result, just like it did before.

```

Comment (Mats-SX) [inline morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala]: This should not throw, but should return an empty result, just like it did before.

```
result.records.collect.toBag shouldBe empty
```

File: morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala [hunk 15]
```diff
@@ -209,13 +224,28 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
...
+      ))
+    }
+
+    it("fails when comparing less than or equal between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val <= b.val2 RETURN a.val"))
```
Mats-SX: Same here; comparing different types is valid in Cypher, but the tuple does not pass the predicate.

morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
```diff
@@ -234,13 +264,27 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
...
+      ))
+    }
+
+    it("fails when comparing greater than between incompatible types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val > b.val2 RETURN a.val"))
```
Mats-SX: And here.

morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
```diff
@@ -260,13 +304,28 @@ class PredicateTests extends MorpheusTestSuite with ScanGraphInit {
...
+      ))
+    }
+
+    it("fails when comparing greater than or equal between different types") {
+
+      // Given
+      val given = initGraph("""CREATE (:A {val: 4})-[:REL]->(:B {val2: 'string'})""")
+
+      // Then
+      a[NoSuitableSignatureForExpr] shouldBe thrownBy(given.cypher("MATCH (a:A)-->(b:B) WHERE a.val >= b.val2 RETURN a.val"))
```
Mats-SX: And here; result should be empty.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -334,96 +343,100 @@ final case class HasType(rel: Expr, relType: RelType) extends PredicateExpressio
...
+    case (CTString, CTString) => Some(CTBoolean)
+    case _ => None
+  }
+}
+
+sealed trait InequalityExprSignature {
+  def signature(lhsType: CypherType, rhsType: CypherType): Option[CypherType] = lhsType -> rhsType match {
+    case (n1, n2) if n1.subTypeOf(CTNumber) && n2.subTypeOf(CTNumber) => Some(CTBoolean)
+    case (c1, c2) if c1.couldBeSameTypeAs(c2) => Some(CTBoolean)
+    case _ => None
```
Mats-SX: This is incorrect (and causes the test failures I commented on above). It is never invalid to compare two values in Cypher, although only values of the same types have a chance of yielding `true`. Therefore we know here that the output type of this will be `CTNull` after the two first cases (or `CTVoid` if these types are expected to be material only).

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -432,6 +445,7 @@ final case class In(lhs: Expr, rhs: Expr) extends BinaryPredicate {
     case _ => childNullPropagatesTo(CTBoolean)
   }
 
+  def signature(lhsType: CypherType, rhsType: CypherType): Option[CypherType] = None //signatures not used as In to special
```
Mats-SX: I think this can be removed; it's all right if this inherits the default from `BinaryPredicate`
```suggestion
```

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -506,16 +520,7 @@ sealed trait ArithmeticExpr extends BinaryExpr {
...
-    val maybeType = signature(lhs.cypherType.material, rhs.cypherType.material)
-
-    maybeType match {
-      case Some(typ) => typ.asNullableAs(rhs.cypherType join lhs.cypherType)
-      case None => throw UnsupportedOperationException(s"Type signature ${getClass.getSimpleName}(${lhs.cypherType}, ${rhs.cypherType}) is not supported.")
-    }
-  }
-
-  def signature(lhsType: CypherType, rhsType: CypherType): Option[CypherType]
+  override def propagationType: Option[PropagationType] = Some(AnyNullable)
```
Mats-SX: I think this one is actually `NullOrAnyNullable` because `node.prop` is `null` when `node` is `null`.

morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
FlorentinD: I took the neo4j implementation as the gold-standard regarding handling InEqualities. 

`
Return 1 < 'a'
Error: org.neo4j.graphdb.QueryExecutionException: 
Type mismatch: expected Float or Integer but was String (line 1, column 20 (offset: 19))
"PROFILE Return 1 < 'a'"`

FlorentinD: But returning an empty result seems also reasonable to me

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: there is only a default propagationType from `BinaryPredicate`.

The signature is not used at all for `In`, but needs to be defined as `BinaryPredicate` extends `TypedExpr`

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
FlorentinD: The file-diff is a bit misleading here.
This propagationType is for `ArithmeticExpr`

morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/PredicateTests.scala
FlorentinD: But to make it consistent the signature of the `InequalityExpression` at the opencypher-frontend should be aligned in the future.

Mats-SX: That's right, the frontend is not consistent here. Unfortunately, Neo4j is also not consistent. Try this query in Neo4j:

```
WITH [1, 'a'] AS l
WITH 1 AS one
  WHERE l[0] < l[1]
RETURN one
```

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
Mats-SX: You're right; I misread the diff.

PR #908, opened 2019-05-03
Shade okapi-relational and relocate transitive dependencies

s1ck:


s1ck:
@tobias-johansson could you do a sanity check? Also, is there a possibility to run the `allJar` task during `[dev]Publish` just for the `okapi-relational-shade` module?

settings.gradle
```diff
@@ -6,6 +6,7 @@ include(':okapi-neo4j-io-testing')
 include(':okapi-ir')
 include(':okapi-logical')
 include(':okapi-relational')
+include(':okapi-relational-shade')
```
mengxr: Not sure if this is the right name. The assembly jar contains:

* o.c.okapi.{relational, logical, ir, trees, impl, api}
* o.c.v9_0.*

mengxr:
@s1ck How to generate the maven pom file? I want to confirm that this fat jar doesn't have any runtime dependencies.

mengxr:
Saw some instructions here: http://imperceptiblethoughts.com/how-to-shadow/#39. Could you give it a try and update the PR? Thanks!

settings.gradle
s1ck: I know. The idea behind that name is coming from the fact that okapi is designed to be an abstraction where you have multiple entry points, i.e.you can implement a non-relational backend that depends on okapi-logical. This leads to the question how e.g. a shaded okapi-logical module would be named if we ever require to have one.

But I see your point. okapi-relational is also the currently most useful entry point and thus it makes sense to name it `okapi-shade` to represent the whole stack.

s1ck:
> @s1ck How to generate the maven pom file? I want to confirm that this fat jar doesn't have any runtime dependencies.

`./gradlew devPublish` generates the maven artifacts and installs them locally. The pom file is located in `okapi-relational-shade/build/publications/dev`

okapi-shade/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+
+apply plugin: 'com.github.johnrengelman.shadow'
+
+description = 'Okapi Shade contains the full Okapi stack and its relocated dependencies'
+
+dependencies {
+    compile project(':okapi-relational')
+}
+
+task allJar(type: ShadowJar) {
```
jjaderberg: Why a custom `type: ShadowJar` task, rather than use the standard `shadowJar` task created by the plugin? Only reason I know to create a custom task is when the standard task is already used for something else in the project, but that's not the case here.

okapi-shade/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+    classifier = 'all'
+    def relocatePackage = 'org.opencypher.relocated.'
```
jjaderberg: The relocation could be made more readable and easier to manage. Something like

```
    def relocatePackage = 'org.opencypher.relocated'
    [
        'cats',
        'fastparse',
        'machinist',
        'org.apache',
        'org.atnos',
        'org.parboiled',
        'sourcecode',
        'ujson',
        'upack',
        'upickle',
    ].each { pkg ->
        relocate(pkg, "${relocatePackage}.$pkg")
    }
```

okapi-shade/build.gradle
s1ck: Good point. We can do the same in `okapi-relational` and `morpheus-spark-cypher`. I guess we need to update a CI task that builds the `allJar` for `morpheus-spark-cypher` and uploads to GitHub?

s1ck: I just saw that we used `shadowJar` before to configure the assembly. @tobias-johansson might know we used a dedicated task.

tobias-johansson: I remember there being a reason, I just can't recall what.. sorry

okapi-shade/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+    dependencies {
+        exclude(dependency('org.scala-lang:'))
+        exclude(dependency('org.scala-lang.modules:'))
+    }
+    exclude "META-INF/versions/**/*"
+    exclude "META-INF/services/**/*"
+    exclude "utf8.json"
+}
+
+pub.dev.artifacts -= 'jar'
```
tobias-johansson: Perhaps just overwrite the whole list here? This sub-project should only publish the `shadowJar` right?

mengxr:
@s1ck We need to correct the published pom file. It shouldn't claim any dependencies except Scala:

~~~
  <dependencies>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-library</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-reflect</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-compiler</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-api</artifactId>
      <version>2.11.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-api-scala_2.12</artifactId>
      <version>11.0</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.opencypher</groupId>
      <artifactId>okapi-relational</artifactId>
      <version>0.3.3-SNAPSHOT</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
~~~

jjaderberg:
@mengxr Why is the pom a problem? Those dependencies should never be resolved beyond the shaded jar.
In this case, since there is no non-shaded jar produced for this module, we can generate the pom based on the shadow configuration instead. This would give us the "dependency-reduced pom", and we can publish that. I don't think it's necessary, but it also wouldn't be a problem, to make that change.

mengxr:
When you publish `okapi-shade` to maven central with a pom that still claims `okapi-relational` and other packages as dependencies. User program will download both `okapi-shade` jar and `okapi-relational` and all its transitive dependencies, against the purpose of shading.

jjaderberg:
@mengxr That depends on how the dependency is resolved--it can be perfectly fine to have those in the pom, and still resolve them from within the shaded jar.

In any case, I changed it to generate a dependency-reduced pom:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.opencypher</groupId>
  <artifactId>okapi-shade</artifactId>
  <version>0.3.4-SNAPSHOT</version>
  <name>Okapi Shade contains the full Okapi stack and its relocated dependencies</name>
  <description>Okapi Shade contains the full Okapi stack and its relocated dependencies</description>
  <url>https://www.opencypher.org</url>
  <licenses>
    <license>
      <name>Apache License, Version 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>morpheus</id>
      <name>The Morpheus team</name>
      <email>opencypher@neo4j.com</email>
      <url>https://www.opencypher.org</url>
    </developer>
  </developers>
  <scm>
    <url>https://github.com/opencypher/morpheus</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-library</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-reflect</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.scala-lang</groupId>
      <artifactId>scala-compiler</artifactId>
      <version>2.12.8</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

PR #910, opened 2019-05-09
AcceptanceTestGenerator generates more compact test

FlorentinD:
Reduces the white-lines between the different steps.

### DIFF
--- a/okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/AcceptanceTestGenerator.scala
+++ b/okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/AcceptanceTestGenerator.scala
@@ -191,8 +191,9 @@ case class AcceptanceTestGenerator(
     val execString = stepsToString(execSteps.zipWithIndex)
     val testString =
       s"""val graph = $initQueryString
-         |    $execString
-       """.stripMargin
+         |$execString
+         """.stripMargin
+    val alignedTestString = alignString(testString, 1)
 
     val expectsError = scenario.steps.exists {
       case _: ExpectError => true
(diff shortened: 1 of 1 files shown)


okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/AcceptanceTestGenerator.scala
```diff
@@ -190,8 +190,7 @@ case class AcceptanceTestGenerator(
 
     val execString = stepsToString(execSteps.zipWithIndex)
     val testString =
-      s"""
-         |    val graph = $initQueryString
+      s"""val graph = $initQueryString
```
Mats-SX: Since you de-indent the `val graph`, why not de-indent the `$execString` as well?

okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/AcceptanceTestGenerator.scala
```diff
@@ -239,8 +238,7 @@ case class AcceptanceTestGenerator(
               case None => ""
             }
             val stepString =
-              s"""
-                 |    lazy val result$stepNr = graph.cypher(
+              s"""lazy val result$stepNr = graph.cypher(
```
Mats-SX: Same thing here, the indentation should probably be consistent on the following lines as well.

okapi-tck/src/main/scala/org/opencypher/okapi/tck/test/AcceptanceTestGenerator.scala
FlorentinD: Good idea, previously I didn't find a way, that wouldnt mess up the format.

The keypoint was adding a `|` even on the first line of the strings.
Bit of a weird behavior, i think.

Mats-SX: Yeah that's how Scala multiline strings work :) Everything between the linebreak and the `|` gets stripped.

PR #919, opened 2019-05-28
Map projection

FlorentinD:
Support of map-projections like `MATCH (n) RETURN n {.* , name:"Bob"}`

### DIFF
--- a/morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
+++ b/morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
@@ -403,14 +403,13 @@ object SparkSQLExprMapper {
           case other => throw IllegalArgumentException("an expression of type CTMap", other)
         }
 
-        case MapProjection(mapOwner, items, includeAllProps) =>
+        case MapProjection(mapOwner, items, includeAllProps) => {
           val convertedItems = items.map { case (key, value) => value.asSparkSQLExpr.as(key) }
-          val itemKeys = items.map(_._1)
+          val itemKeys = items.map { case (propKey, _) => propKey }
           val intersectedMapItems = if (includeAllProps) {
             mapOwner.cypherType.material match {
               case x if x.subTypeOf(CTElement) =>
-                val uniqueEntityProps = header.propertiesFor(mapOwner)
-                  .filterNot(p => itemKeys.contains(p.key.name))
+                val uniqueEntityProps = header.propertiesFor(mapOwner).filterNot(p => itemKeys.contains(p.key.name))
                 val propertyColumns = uniqueEntityProps.map(p => p.asSparkSQLExpr.as(p.key.name))
                 convertedItems ++ propertyColumns
               case CTMap(inner) =>
@@ -422,6 +421,7 @@ object SparkSQLExprMapper {
           else convertedItems
 
           create_struct(intersectedMapItems)
+        }
 
         // Aggregators
         case Count(_, distinct) =>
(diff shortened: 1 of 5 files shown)


morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
```diff
@@ -404,6 +403,26 @@ object SparkSQLExprMapper {
           case other => throw IllegalArgumentException("an expression of type CTMap", other)
         }
 
+        case MapProjection(mapOwner, items, includeAllProps) =>
+          val convertedItems = items.map { case (key, value) => value.asSparkSQLExpr.as(key) }
+          val itemKeys = items.map(_._1)
```
s1ck: We try to avoid the `_._1` syntax for tuple element access. It's less readable compared to `.map { case (propKey, _) => ... }`

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
```diff
@@ -404,6 +403,26 @@ object SparkSQLExprMapper {
...
+                  .filterNot(p => itemKeys.contains(p.key.name))
+                val propertyColumns = uniqueEntityProps.map(p => p.asSparkSQLExpr.as(p.key.name))
+                convertedItems ++ propertyColumns
+              case CTMap(inner) =>
+                val uniqueMapKeys = inner.keys.filterNot(key => itemKeys.contains(key))
+                val uniqueMapColumns = uniqueMapKeys.map(key => child0.getField(key).as(key))
+                convertedItems ++ uniqueMapColumns
+            }
+          }
+          else convertedItems
```
s1ck: Wrap in curly braces

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
```diff
@@ -404,6 +403,26 @@ object SparkSQLExprMapper {
...
         }
 
+        case MapProjection(mapOwner, items, includeAllProps) =>
+          val convertedItems = items.map { case (key, value) => value.asSparkSQLExpr.as(key) }
+          val itemKeys = items.map(_._1)
+          val intersectedMapItems = if (includeAllProps) {
+            mapOwner.cypherType.material match {
+              case x if x.subTypeOf(CTElement) =>
+                val uniqueEntityProps = header.propertiesFor(mapOwner)
+                  .filterNot(p => itemKeys.contains(p.key.name))
```
s1ck: should fit on previous line?

FlorentinD: thought that would be a better style, but I guess as `propertiesFor()` is no complex function I will inline it.

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
FlorentinD: I somehow keep forgetting to do so :sweat_smile:

s1ck:
Can you do another rebase? We merged a fix for the flaky Hive tests, would be great to have this build green, too.

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/impl/SparkSQLExprMapper.scala
s1ck: @FlorentinD I think you missed that one

FlorentinD: Seems like you merged it nevertheless.
Thought you meant the wohle block with your comment

PR #922, opened 2019-06-03
Fix flaky Hive tests

s1ck:
The main issue was that the `Customer360Example` did not use the correct Hive configuration when instantiating the Spark Session. Instead of using Derby (`metastore_db`) in-memory, it created a persistent version of it which lead to race conditions between our examples during CI execution trying to create a metastore.

We fixed this by aligning Spark configuration of `TestSparkSession` and `MorpheusSession.local()` and set the correct derby settings by default.


morpheus-examples/src/main/scala/org/opencypher/morpheus/examples/CustomDataFrameInputExample.scala
```diff
@@ -43,12 +43,8 @@ object CustomDataFrameInputExample extends App {
 
   // 1) Create Morpheus session and retrieve Spark session
   // tag::create-session[]
-  val spark: SparkSession = SparkSession
```
jjaderberg: It's useful to show the user how to use an existing spark session, so prefer not to change this

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/MorpheusSession.scala
```diff
@@ -93,19 +94,36 @@ object MorpheusSession extends Serializable {
...
+  val localSparkConf: SparkConf = {
     val conf = new SparkConf(true)
     conf.set("spark.sql.codegen.wholeStage", "true")
     conf.set("spark.sql.shuffle.partitions", "12")
     conf.set("spark.default.parallelism", "8")
-    conf.setAll(settings)
+    // Required for left outer join without join expressions in OPTIONAL MATCH (leads to cartesian product)
+    conf.set("spark.sql.crossJoin.enabled", "true")
+
+    //    setting for debug
```
jjaderberg: These comments are not pretty :/ Can we move them (you suggest SparkSessionFixture)

morpheus-examples/src/main/scala/org/opencypher/morpheus/examples/CustomDataFrameInputExample.scala
s1ck: Good point, it's the only example that demonstrates using a custom spark session.

PR #924, opened 2019-06-04
Add test for incompatible graphs in CONSTRUCT

FlorentinD:


morpheus-testing/src/test/scala/org/opencypher/morpheus/impl/acceptance/MultipleGraphTests.scala
```diff
@@ -972,4 +973,13 @@ class MultipleGraphTests extends MorpheusTestSuite with ScanGraphInit {
       CypherMap("n" -> MorpheusNode(0, Set("FOO")))
     )
   }
+
+  it("Fail early on incompatible graphs in CONSTRUCT") {
```
DarthMax:
```suggestion
  it("fails early on incompatible graphs in CONSTRUCT") {
```

PR #925, opened 2019-06-05
Restrict possible patterns in Construct

FlorentinD:
cloned relationships should not be allowed in the patterns, with the matched source and target node.

The solution checks the relationship patterns in the CONSTRUCT-clauses by matching against propagated MATCH-patterns.

### DIFF
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/util/FreshVariableNamer.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/util/FreshVariableNamer.scala
@@ -31,8 +31,12 @@ import org.opencypher.okapi.ir.api.expr.Var
 
 object FreshVariableNamer {
   val PREFIX = "  "
+  val NAME = "FRESH_VAR"
+
+  def notNamed(x: String): Boolean = x.startsWith(PREFIX + NAME)
+  def isNamed(x: String): Boolean = !notNamed(x)
 
   def apply(seed: String, t: CypherType): Var = Var(s"$PREFIX$seed")(t)
 
-  def apply(id: Int, t: CypherType): Var = apply(s"FRESH_VAR$id", t)
+  def apply(id: Int, t: CypherType): Var = apply(s"$NAME$id", t)
 }
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -37,14 +37,14 @@ import org.opencypher.okapi.impl.exception.{IllegalArgumentException, IllegalSta
 import org.opencypher.okapi.ir.api._
 import org.opencypher.okapi.ir.api.block.{SortItem, _}
 import org.opencypher.okapi.ir.api.expr._
-import org.opencypher.okapi.ir.api.pattern.Pattern
+import org.opencypher.okapi.ir.api.pattern.{Connection, Pattern}
 import org.opencypher.okapi.ir.api.set.{SetItem, SetLabelItem, SetPropertyItem}
-import org.opencypher.okapi.ir.api.util.CompilationStage
+import org.opencypher.okapi.ir.api.util.{CompilationStage, FreshVariableNamer}
 import org.opencypher.okapi.ir.impl.exception.ParsingException
 import org.opencypher.okapi.ir.impl.refactor.instances._
 import org.opencypher.okapi.ir.impl.util.VarConverters.RichIrField
 import org.opencypher.v9_0.ast.QueryPart
-import org.opencypher.v9_0.util.InputPosition
+import org.opencypher.v9_0.util.{FreshIdNameGenerator, InputPosition}
 import org.opencypher.v9_0.{ast, expressions => exp}
 
 
@@ -308,6 +308,21 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
 
             // Fields inside of CONSTRUCT could have been matched on other graphs than just the workingGraph
             val cloneSchema = schemaForElementTypes(context, cloneItemMap.values.map(_.cypherType).toSet)
+
+            //Make sure cloned relationship are in no pattern with unnamed nodes
+            cloneItemMap.values.foreach {
+              case r@RelationshipVar(relName) =>
+                val illegalRelPattern = createPatterns.filter(pattern =>
+                  pattern.topology.exists {
+                    case (IRField(fieldName), c: Connection) if fieldName == relName =>
+                      c.endpoints.exists(node => FreshVariableNamer.notNamed(node.name))
+                    case _ => false
+                  }
+                )
+                if (illegalRelPattern.nonEmpty)
+                  throw UnsupportedOperationException(s"Cloned relationships are not allowed in the same pattern as anonymous nodes. \n Found in ${illegalRelPattern.head}")
+              case _ => ()
+            }
 
             // Make sure that there are no dangling relationships
             // we can currently only clone relationships that are also part of a new pattern
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -39,12 +39,12 @@ import org.opencypher.okapi.ir.api.block.{SortItem, _}
 import org.opencypher.okapi.ir.api.expr._
 import org.opencypher.okapi.ir.api.pattern.{Connection, Pattern}
 import org.opencypher.okapi.ir.api.set.{SetItem, SetLabelItem, SetPropertyItem}
-import org.opencypher.okapi.ir.api.util.{CompilationStage, FreshVariableNamer}
+import org.opencypher.okapi.ir.api.util.CompilationStage
 import org.opencypher.okapi.ir.impl.exception.ParsingException
 import org.opencypher.okapi.ir.impl.refactor.instances._
 import org.opencypher.okapi.ir.impl.util.VarConverters.RichIrField
 import org.opencypher.v9_0.ast.QueryPart
-import org.opencypher.v9_0.util.{FreshIdNameGenerator, InputPosition}
+import org.opencypher.v9_0.util.InputPosition
 import org.opencypher.v9_0.{ast, expressions => exp}
 
 
@@ -317,9 +317,10 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
                   pattern.topology.foreach {
                     case (relField@IRField(fieldName), con: Connection) if fieldName == relName =>
                       val expectedConnection = context.knownPatterns.map(_.topology).flatMap(_.get(relField)).headOption
-                        .getOrElse(throw UnsupportedOperationException(s"Cloned relationship '$fieldName' needs a match pattern to resolve the src and target nodes (maybe out of scope due f.i. WITH-Clauses)."))
+                        .getOrElse(throw IllegalArgumentException(s"Cloned relationship '$fieldName' needs its source and target nodes from the MATCH  pattern (could be out of scope after a WITH)."))
                       if (expectedConnection != con)
-                        throw UnsupportedOperationException(s"Cloned relationships are are only allowed with the corresponding source and target node. \n Found $con \n expected $expectedConnection")
+                        throw IllegalArgumentException(s"Cloned relationships are are only allowed with the corresponding source and target node", con
+                          , s"Following pattern was expected: \n $expectedConnection")
                     case _ => ()
                   }
                 }
@@ -373,7 +374,11 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
               createPattern,
               setItems,
               onGraphs)
-            val updatedContext = context.resetKnownPatterns.withWorkingGraph(patternGraph).registerSchema(qgn, patternGraphSchema)
+
+            val updatedContext = context.resetKnownPatterns
+              .withWorkingGraph(patternGraph)
+              .registerSchema(qgn, patternGraphSchema)
+              .resetKnownTypes
             put[R, IRBuilderContext](updatedContext) >> pure[R, List[Block]](List.empty)
           }
         } yield refs
--- a/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
+++ b/okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
@@ -214,7 +214,7 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
 
             val typedOutputs = typedMatchBlock.outputs(block)
             val updatedRegistry = blockRegistry.register(block)
-            val updatedContext = context.withBlocks(updatedRegistry).withFields(typedOutputs)
+            val updatedContext = context.withBlocks(updatedRegistry).withFields(typedOutputs).registerPattern(pattern)
             put[R, IRBuilderContext](updatedContext) >> pure[R, List[Block]](List(block))
           }
         } yield blocks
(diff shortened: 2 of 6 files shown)


okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/util/FreshVariableNamer.scala
```diff
@@ -29,10 +29,14 @@ package org.opencypher.okapi.ir.api.util
 import org.opencypher.okapi.api.types.CypherType
 import org.opencypher.okapi.ir.api.expr.Var
 
+import scala.util.matching.Regex
+
 object FreshVariableNamer {
   val PREFIX = "  "
+  val FRESHVAR = "FRESH_VAR"
```
Mats-SX: Perhaps call this `NAME`?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +309,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
             // Fields inside of CONSTRUCT could have been matched on other graphs than just the workingGraph
             val cloneSchema = schemaForElementTypes(context, cloneItemMap.values.map(_.cypherType).toSet)
 
+            //Make sure cloned relationship are in no pattern with unnamed nodes
+            cloneItemMap.foreach {
+              case (_, RelationshipVar(relName)) =>
```
Mats-SX: We could do 

```
cloneItemMap.values.foreach {
  case RelationshipVar(relName) =>
  ...
```

to avoid the tuple where we just ignore the first value.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +309,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
+                val illegalRelPattern = createPatterns.filter(pattern =>
+                  pattern.fields.exists {
+                    case field@IRField(fieldName) if field.cypherType.couldBeSameTypeAs(CTRelationship) =>
```
Mats-SX: We could iterate over `pattern.rels` instead, something like:

```
pattern.rels.exists { rel =>
  val matchesRel = rel.name == relName
  ...
}
```

This avoids the guard in the `case` that checks for relationships.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +309,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
+                      val matchesRel = fieldName.equals(relName)
```
Mats-SX: In Scala we prefer `==` over `.equals()` :)

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +309,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
+                      val unnamedPart = pattern.fields.exists(f => f.name.matches(FreshVariableNamer.freshVarPattern))
```
Mats-SX: This probably works all right, but I think we can do something a little bit better. Since we have processed the incoming pattern _just before_ we do this analysis, I think it would make more sense that we _mark_ the auto-named variables at naming time. We could model this by adding a field to the `Var` class called `isNameGenerated` or `anonymous` or something similar. The default value would be `false`, but for variables created by the `FreshVariableNamer` it would be set to `true`. That would allow us to simply check the value of that field here, instead of relying on string matching. What do you think?

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +309,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
+                      matchesRel && unnamedPart
+                    case _ => false
+                  }
+                )
+                if (illegalRelPattern.nonEmpty)
+                  throw UnsupportedOperationException(s"Cloned relations are not allowed in the same pattern as anonymous nodes. \n Found in ${illegalRelPattern.head}")
```
Mats-SX:
```suggestion
                  throw UnsupportedOperationException(s"Cloned relationships are not allowed in the same pattern as anonymous nodes. \n Found in ${illegalRelPattern.head}")
```

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -691,33 +691,43 @@ class IrBuilderTest extends IrTestSuite {
...
+      an[UnsupportedOperationException] shouldBe thrownBy(query.asCypherQuery().model.result)
+    }
+
+    //as the relationship-var does not contain any info about src and trg
+    //maybe solve via push-down of match patterns in IR ?
+    ignore("fails cloning rel and cloning false src") {
+      val query =
+        """|MATCH (a)-[e]->(b)
+           |CONSTRUCT
+           |  CREATE (b)-[e]->(a)
```
Mats-SX: I think we can not sensible allow this operation. So even if variables are not auto-named, I think we must make this fail. The one case that we could possibly allow would be

```
MATCH (a)-[r]->(b)
CONSTRUCT
  CREATE (a)-[r]->(b)
```

but I'm not sure our pattern representation can support it. Perhaps if we analyse the `Connection` objects somehow? But what makes it even harder are queries like this:

```
MATCH (a)-[r]->(b)
WITH r
MATCH (a)
MATCH (b)
CONSTRUCT
  CREATE (a)-[r]->(b) // not valid!
```

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
FlorentinD: My approach also seemed a bit hackish to me. 
I also think that a boolean check is better readable than the string matching

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
FlorentinD: It won't be easy to push down the patterns. Your last example could get included I think (if there are only a limited number of corner case pattern

So another possibility would be to restrict cloning of relationships at all and suggest to use COPY OF instead. (I think there is already a comment in the trello card)

FlorentinD: @mats I will try introducing a match-pattern scope and WITH, RETURN and CONSTRUCT would clear the current scope.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
FlorentinD: The problem is that at this point we don't have `Vars`, but `IRFields`, therefore I only can match the name.
But while trying, I realized that the front-end has a `FreshIdNameGenerator`, making our own `FreshVariableNamer` obsolete (?).

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +310,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
             val cloneSchema = schemaForElementTypes(context, cloneItemMap.values.map(_.cypherType).toSet)
 
+            //Make sure cloned relationships have the correct source and target nodes
+            cloneItemMap.values.foreach {
+              case RelationshipVar(relName) =>
+                createPatterns.foreach { pattern => {
+                  pattern.topology.foreach {
+                    case (relField@IRField(fieldName), con: Connection) if fieldName == relName =>
+                      val expectedConnection = context.knownPatterns.map(_.topology).flatMap(_.get(relField)).headOption
+                        .getOrElse(throw UnsupportedOperationException(s"Cloned relationship '$fieldName' needs a match pattern to resolve the src and target nodes (maybe out of scope due f.i. WITH-Clauses)."))
```
Mats-SX:
```suggestion
                        .getOrElse(throw UnsupportedOperationException(s"Cloned relationship '$fieldName' needs its source and target nodes from the MATCH pattern (could be out of scope after a WITH)."))
```

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
```diff
@@ -309,6 +310,23 @@ object IRBuilder extends CompilationStage[ast.Statement, CypherStatement, IRBuil
...
+                      if (expectedConnection != con)
+                        throw UnsupportedOperationException(s"Cloned relationships are are only allowed with the corresponding source and target node. \n Found $con \n expected $expectedConnection")
```
Mats-SX: I think we could use `org.opencypher.okapi.impl.exception.IllegalArgumentException` which provides a nice "expected"/"found" format already.

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
Mats-SX: I don't think the exception type is the best; perhaps we could use `IllegalArgumentException` instead?

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -691,33 +691,134 @@ class IrBuilderTest extends IrTestSuite {
...
+    it("CONSTRUCT clears pattern scope") {
+      val query =
+        """|MATCH (a)-[e]->(b)
+           |CONSTRUCT
+           |  CREATE (a)-[:TYP]->()
+           |CONSTRUCT
+           |  CREATE (a)-[e]->(b)
+           |RETURN GRAPH""".stripMargin
+
+      an[UnsupportedOperationException] shouldBe thrownBy(query.asCypherQuery().model.result)
```
Mats-SX: This isn't correct; `CONSTRUCT` clears the scope between blocks, so none of `a`, `e`, `b` are in scope at the second `CONSTRUCT` (as the first cleared the entire scope), meaning the pattern `CREATE (a)-[e]->(b)` is valid and will create a new relationship and two new nodes (no cloning going on as all three vars are fresh).

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -691,33 +691,134 @@ class IrBuilderTest extends IrTestSuite {
...
+    it("WITH clears patterns scope") {
+      val query =
+        """|MATCH (a)-[e]->(b)
+           |WITH e
+           |CONSTRUCT
+           |  CREATE (a)-[e]->(b)
+           |RETURN GRAPH""".stripMargin
+
+      an[UnsupportedOperationException] shouldBe thrownBy(query.asCypherQuery().model.result)
+    }
```
Mats-SX: Could we also add a slightly more malevolent query like this:

```
        """|MATCH (a)-[e]->(b)
           |WITH e, b AS a, a AS b
           |CONSTRUCT
           |  CREATE (a)-[e]->(b)
           |RETURN GRAPH""".stripMargin
```

It should also fail, but I'm not able to determine whether it will currently :)

FlorentinD: I would say the frontend should annotate the `a` with its positions and therefore these are different ones and it should fail.

But I will try it out :)

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
FlorentinD: Right `Construct` should clear the scope, but I think there is a bug in the frontend regarding this.
I will look further into it

okapi-ir/src/main/scala/org/opencypher/okapi/ir/impl/IRBuilder.scala
FlorentinD: Ah right I switched between those two in the process

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
FlorentinD: I could clear the scope for the following CONSTRUCT, but not for following RETURN blocks (would need changes in the front-end).

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
FlorentinD: works as thought :+1:

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
Mats-SX: I understand. Well let's leave that open as a separate fix. It's not going to matter for this feature because we can not do any cloning in `RETURN` anyway.

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
Mats-SX: Perfect

okapi-ir/src/test/scala/org/opencypher/okapi/ir/impl/IrBuilderTest.scala
```diff
@@ -691,7 +691,7 @@ class IrBuilderTest extends IrTestSuite {
           |RETURN GRAPH
         """.stripMargin
 
-       an[UnsupportedOperationException] shouldBe thrownBy(query.asCypherQuery().model.result)
+       an[IllegalArgumentException] shouldBe thrownBy(query.asCypherQuery().model.result)
```
Mats-SX: The reason CI fails is because this is using `java.lang.IllegalArgumentException` as `java.lang` requires no import. Unfortunately you need to explicitly import our `IllegalArgumentException` to override the implicit import.

FlorentinD: Thank you.
I was working on the WITH clause and fixed this also with the following commit.

Looking forward to your review of my latest changes :smile:

Mats-SX:
I think this looks good to go in if we can just add a few acceptance-level tests to increase confidence that it works as intended when cloning is allowed.

FlorentinD:
You had the right feeling. 
Aliasing cloned variables inside CONSTRUCT wasn't considered before my last commit.

jjaderberg:
Could you add acceptance tests in MultipleGraphTests, something like

```
  it("should clone a relationship") {
    val query =
      """|MATCH (a)-[r:HAS_SIMILAR_NAME]->(b)
         |CONSTRUCT
         |  CREATE (a)-[r]->(b)
         |RETURN GRAPH""".stripMargin
    val result = testGraphRels.cypher(query)
    result.graph.cypher("MATCH ()-[r]->() RETURN type(r) as type").records.iterator.toBag should equal(Bag(
      CypherMap("type" -> "HAS_SIMILAR_NAME")
    ))
  }
```

FlorentinD:
@jjaderberg I didn't add one yet, as my changes only work on the IR lvl.
I'm surprised there are only test-cases for explicit cloning.

I will add some tomorrow :+1: .
I assume it is enough to only add positive ones here, as shouldn't add anything new as the errors are thrown at IR-lvl.

PR #929, opened 2019-06-25
Fix PGDS id generation

FlorentinD:
serialized relationship ids are not based on property columns anymore, as this lead to errors with unsupported types for serialization.
Instead a monotonically_increasing_id() is used for unique ids between multiple relationships.

### DIFF
--- a/morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
+++ b/morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
@@ -28,6 +28,7 @@ package org.opencypher.morpheus.api.io.sql
 
 import java.net.URI
 
+import org.apache.spark.sql.functions.monotonically_increasing_id
 import org.apache.spark.sql.types.{DataType, DecimalType, IntegerType, LongType}
 import org.apache.spark.sql.{Column, DataFrame, DataFrameReader, functions}
 import org.opencypher.graphddl._
@@ -219,8 +220,9 @@ case class SqlPropertyGraphDataSource(
     evm: EdgeToViewMapping,
     df: DataFrame
   ): (Map[String, String], Seq[Column]) = {
-
-    val relIdColumn = generateIdColumn(df, evm.key, df.columns.map(_.decodeSpecialCharacters).toList, relSourceIdKey, schema)
+    val genIdColumns = df.columns.filterNot(col => evm.propertyMappings.contains(col))
+      .map(_.decodeSpecialCharacters)
+    val relIdColumn = generateIdColumn(df, evm.key, genIdColumns.toList, relSourceIdKey, schema, relId = true)
     val relSourceIdColumn = generateIdColumn(df, evm.startNode.nodeViewKey, evm.startNode.joinPredicates.map(_.edgeColumn), sourceStartNodeKey, schema)
     val relTargetIdColumn = generateIdColumn(df, evm.endNode.nodeViewKey, evm.endNode.joinPredicates.map(_.edgeColumn), sourceEndNodeKey, schema)
     val relProperties = generatePropertyColumns(evm, df, ddlGraph, schema, Some("relationship"))
@@ -349,7 +351,8 @@ case class SqlPropertyGraphDataSource(
     elementViewKey: ElementViewKey,
     idColumnNames: List[String],
     newIdColumn: String,
-    schema: PropertyGraphSchema
+    schema: PropertyGraphSchema,
+    relId: Boolean = false
   ): Column = {
     val idColumns = idColumnNames.map(_.toLowerCase.encodeSpecialCharacters).map(dataFrame.col)
     idGenerationStrategy match {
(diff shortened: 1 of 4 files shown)


FlorentinD:
So parallel edges with same type and no keys already get different id's even without adding an monocally_increasing_id() column.

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -28,6 +28,7 @@ package org.opencypher.morpheus.api.io.sql
 
 import java.net.URI
 
+import org.apache.spark.sql.functions.monotonically_increasing_id
```
s1ck: Where is the udf being used?

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -219,8 +220,9 @@ case class SqlPropertyGraphDataSource(
     evm: EdgeToViewMapping,
     df: DataFrame
   ): (Map[String, String], Seq[Column]) = {
-
-    val relIdColumn = generateIdColumn(df, evm.key, df.columns.map(_.decodeSpecialCharacters).toList, relSourceIdKey, schema)
+    val genIdColumns = df.columns.filterNot(col => evm.propertyMappings.contains(col))
```
s1ck: Does the DF already contain a unique id for each row from some pre-processing step? Otherwise I can't imagine how this works for parallel edges.

okapi-testing/src/main/scala/org/opencypher/okapi/testing/PGDSAcceptanceTest.scala
```diff
@@ -252,6 +261,12 @@ trait PGDSAcceptanceTest[Session <: CypherSession, Graph <: PropertyGraph] {
         session.catalog.source(ns).graph(g3).relationships("r").size shouldBe 0
       },
 
+      Scenario("API: PropertyGraphDataSource: unique IDs for graph #4", g5) { implicit ctx: TestContext =>
```
s1ck:
```suggestion
      Scenario("API: PropertyGraphDataSource: unique IDs for graph #5", g5) { implicit ctx: TestContext =>
```

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
FlorentinD: See my comment above, I used it previously.
I couldn't figure out yet, why they are still unique without.
I will try figuring it out on Sunday :)

s1ck: Sorry didn't see the comment. Interesting, I'll have a look, too.

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
FlorentinD: yes there already is a unique id column

morpheus-spark-cypher/src/main/scala/org/opencypher/morpheus/api/io/sql/SqlPropertyGraphDataSource.scala
```diff
@@ -220,9 +219,12 @@ case class SqlPropertyGraphDataSource(
     evm: EdgeToViewMapping,
     df: DataFrame
   ): (Map[String, String], Seq[Column]) = {
-    val genIdColumns = df.columns.filterNot(col => evm.propertyMappings.contains(col))
-      .map(_.decodeSpecialCharacters)
-    val relIdColumn = generateIdColumn(df, evm.key, genIdColumns.toList, relSourceIdKey, schema, relId = true)
+
+    val relIdColumn = {
+      val inputIdColumn = df.columns.find(col => col == "id").map(_.decodeSpecialCharacters)
```
s1ck: You could use `df.col(SourceIdKey.name)` instead of the `find` method and inline into the `generateIdColumn` call.

FlorentinD: ah ok, I was looking for something like this :+1:

FlorentinD: `SourceIdKey.name` is enough, as I need the column name here.

FlorentinD: Seems like in `SqlPropertyGraphDataSourceTest` are some tests, where we do not have an id column.
Therefore I will reintroduce the monocally_increasing_id column for the id generation

FlorentinD: Although they seem to get different ids, without having any id-columns.
Will look into it further

FlorentinD: Scenario: `reads relationships from multiple tables` leads to the same id when using `monotonically_increasing_id`

Issue #936, opened 2019-10-07
Converting Milliseconds GMT Date to years and Months Issue


MohamedRagabAnas:
Hi, There 
I'm wondering what is the best way to take the year and month from the Milliseconds GMT Date ( 1313591219961) in Morpheus Cypher. 
I'm using basic calculations to get the year and month from this date, it works fine in neo4j, but this gives me some error with Morpheus Cypher!
   
      |MATCH
      |  (home:Country{name:'Germany'})<-[:IS_PART_OF]-(:City)<-[:IS_LOCATED_IN]-
      |  (:Person)<-[:HAS_CREATOR]-(message:Message)-[:IS_LOCATED_IN]->(destination:Country)
      |WHERE home <> destination
      |WITH
      |  message,
      |  destination,
      |  ( (message.creationDate%31556952000) /2592000000 +1 ) AS month
      |RETURN
      |  count(message) AS messageCount,
      |  destination.name,
      |  month
      |ORDER BY
      |  messageCount DESC,
      |  destination.name ASC,
      |  month ASC
      |LIMIT 100
    
Error shown is: 

`Exception in thread "main" org.opencypher.okapi.impl.exception.NotImplementedException: Not yet able to convert expression: Modulo(Property(Variable(message),PropertyKeyName(creationDate)),Parameter(  AUTOINT1,Integer))`

Thanks in advance for your help!

Mats-SX:
The reason you get this error is because the modulo operator (`%`) is not yet implemented. If you try to do your calculations without it, perhaps you can work around this limitation?

It would probably be easy to add this function, however we are not actively looking at additional features for Morpheus currently.

Mats-SX:
The modulo operator expression (`%`) has now been implemented.

PR #939, opened 2019-10-11
Implement modulo expression

DarthMax:


Mats-SX:
Fixes #936

okapi-ir/src/main/scala/org/opencypher/okapi/ir/api/expr/Expr.scala
```diff
@@ -611,6 +611,20 @@ final case class Divide(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
...
+final case class Modulo(lhs: Expr, rhs: Expr) extends ArithmeticExpr {
+
+  override val op = "%"
+
+  override def signature(lhsType: CypherType, rhsType: CypherType): Option[CypherType] = lhsType -> rhsType match {
+    case (CTVoid, _) | (_, CTVoid) => Some(CTNull)
+    case (CTInteger, CTInteger) => Some(CTInteger)
+    case (CTFloat, CTFloat) => Some(CTFloat)
+    case (CTInteger, CTFloat) => Some(CTFloat)
+    case (CTFloat, CTInteger) => Some(CTFloat)
```
Mats-SX: Is the output not always `CTInteger` actually? The modulo would never include fractions AFAIK?

DarthMax: Actually if you have `2.9 % 2.8` it is `0.1` :)

PR #983, opened 2025-11-25
Use Scalafmt code formatter

loveleif:
Makes it a bit easier to work with the repo when fixing things in scala.


michael-simons (approved):
After carefully reading all 37012 changed lines, I think that looks really good.

(I did the same in a couple of projects, in the end it's much nicer…)

